PR #2, opened 2020-07-20
Fix build on linux

aosync:
glfw for some reason requires X11 and Xxf86vm to be explicitly marked before it.


AlexandreRouma:
Since I changed a bunch of stuff in the CMakeFiles.txt, I can't merge the request
But I added the two lines you suggested to the new version.

Thank you for your contribution.

PR #16, opened 2020-09-06
Freq can be zero

howard0su:
In Zero-IF scenerio, freq can be zero.


AlexandreRouma:
Thanks for contributing, I'll be merging your pull request and add you to the contributor's list

PR #19, opened 2020-09-06
Better plugin build

howard0su:
1. reduce amount of duplicate cmake scripts
2. Put the plugin dll into the same folder of .exe.
3. Omit the abs path for plugin to avoid complexity when testing debug/release build.


AlexandreRouma:
Hi,
Thanks for contributing, but unfortunately, the way plugins are built is because they're meant to be easy to build on their own without requiring the main CMakeLists.

Also, the files are checked in the code because the module_list file is meant to be modified by the user to add/remove modules. Not checking the existance and validity of the file will result in a crash with no explanation of what happened.

And finally, the path in the module list has been changed in later revisions where the linux build has a path that leads to /etc/sdrpp
and the windows build has a path that's just ./
This way, modules can be easily loaded from anywhere if the user desires to do so.

I'm sorry but I don't think I can merge this pull request.
Cheers, alex.

PR #28, opened 2020-10-22
Build system fix and cleanup

howard0su:
Couple of fixes to the build system:
1. Fix build on Debian 10
2. Use pkg-config to find out all the path of includes/libs on Linux and OSX, and remove all hardcoded path.


aosync:
This is gonna be helpful for OpenBSD too.

PR #42, opened 2020-12-14
Soapysdr: Improve sample rate information

wingrime:
- Add "Sample Rate" text
- Display sample rate with kHz and MHz when possible


AlexandreRouma:
Hi,

Thanks for contributing!

I'll merge your pull request but keep the current refresh button location (without the "Sample Rate" text)
I think the location and content of the dropdown is self explanatory and the text only "bloats" up the menu.

Best regards,
Alex.

PR #44, opened 2020-12-14
Add all default modules to example

cropinghigh:
And fix soapy -> soapy_source


AlexandreRouma:
this file is user specific is will soon no longer be shipped with the software

PR #76, opened 2021-02-23
Support selecting antenna for SoapySDR source

### LINKED ISSUE #11: Add antenna switching.
SDR's such as the SDRPlay RSP2 and Duo have more than one antenna input.
ability to set the input would be handy.

thotypous:
Partially fixes #11

image

### DIFF
--- a/soapy_source/src/main.cpp
+++ b/soapy_source/src/main.cpp
@@ -1,3 +1,4 @@
+#include <SoapySDR/Constants.h>
 #include <imgui.h>
 #include <spdlog/spdlog.h>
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
Hello!
Thanks for contributing :)

In what way is this only "partial" support? It seems to be doing everything it's supposed to do for antenna switching.
Do you mean channel switching as well?

racerxdl:
I think its because the issue #11 also mentions multi-stage gain and multi-channel.

racerxdl (approved):

PR #85, opened 2021-03-22
Fix file source rewind and read after it is finished

nonoo:
Calling clear() clears the eof state flag which prevents further reading from the stream.


AlexandreRouma:
Thanks for your contribution!

PR #97, opened 2021-04-18
Rewrite rx888 source on libsddc

kpd200081:
Additionally:
 - Show Linux dll load error
 - Add new GUI element - slider like combobox

I would use this program for RX888 mk2 on Linux (because this is my primary OS).
There is a cross-platform library for this sdr and friends: https://github.com/ik1xpv/ExtIO_sddc.
But the Linux part on the master is not very useful, so I decided to improve it and use here instead Windows-only driver (pull request with my current version of library https://github.com/ik1xpv/ExtIO_sddc/pull/196).


kpd200081:
This is still work in progress. But I try to rebase to master andand run into the waterfall and demodulators not handling the stream correctly (something with FFT, I think). So, I want to ask you @Ryzerth: were there any changes in the way of streaming from source?

AlexandreRouma:
Hello,
Yes, a lot has changed since the last time the RX888 source was edited. I recommend completely scrapping the code and starting from scratch (take a look at the airspy source for reference). Do note that some of the code needed is currently broken and needs rewriting (the decimator for example)

PS: If this is gonna be written with libsddc, I recommend writing a general library and not just for the RX888.

Regards,
Alex.

kpd200081:
>Yes, a lot has changed since the last time the RX888 source was edited. I recommend completely scrapping the code and starting from scratch
Of course, I rewrite most of the code. When I started there were only module class and callback, later I return real to I/Q converter, because it works better than what is available in libsddc.
I was looking it rtl-sdr for inspiration. When I ask about changes I mean in last 20 days (where base of my branch), sorry to confusing you.

>If this is gonna be written with libsddc, I recommend writing a general library and not just for the RX888

Of course, RX888 now here just for name, later I should rename it, I think...

AlexandreRouma:
> When I ask about changes I mean in last 20 days (where base of my branch)
Yes, a lot is changing. Today for example I just fixed all cmake files to allow building on OSX

kpd200081:
Oh well, I'll look in other modules to see what I need to change. 
Thanks for the reply anyway.

CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ option(OPT_BUILD_SDRPLAY_SOURCE "Build SDRplay Source Module (Depedencies: libsd
 option(OPT_BUILD_PLUTOSDR_SOURCE "Build PlutoSDR Source Module (Depedencies: libiio, libad9361)" ON)
 option(OPT_BUILD_HACKRF_SOURCE "Build HackRF Source Module (Depedencies: libhackrf)" ON)
 option(OPT_BUILD_RTL_SDR_SOURCE "Build RTL-SDR Source Module (Depedencies: librtlsdr)" ON)
+option(OPT_BUILD_RX888_SOURCE "Build RX-888 Source Module (Depedencies: libsddc)" ON)
```
AlexandreRouma, line 21: Wrong dependency listed, should be `libusb-1.0`

core/src/gui/widgets/combo_as_slider.cpp
```diff
@@ -0,0 +1,30 @@
...
+namespace ImGui {
+    bool ComboAsSlider(const char* label, int* current_item, const char* items_separated_by_zeros){
+        int items_count = 0;
+        const char* p = items_separated_by_zeros;
+        while (*p)
+        {
+            p += strlen(p) + 1;
+            items_count++;
+        }
+        char** items_starts = new char*[items_count];
```
AlexandreRouma, line 15: Heap allocation and deallocation on every frame is VERY BAD for performance. Also code style from contributing.md isn't respected.

CMakeLists.txt
kpd200081: Thanks for advice

It's anyway far from master, I will try to find a time to rebase

core/src/gui/widgets/waterfall.cpp
```diff
@@ -507,7 +506,7 @@ namespace ImGui {
                 doZoom(drawDataStart, drawDataSize, dataWidth, &rawFFTs[((i + currentFFTLine) % waterfallHeight) * rawFFTSize], tempData, _fastFFT);
                 for (int j = 0; j < dataWidth; j++) {
                     pixel = (std::clamp<float>(tempData[j], waterfallMin, waterfallMax) - waterfallMin) / dataRange;
-                    waterfallFb[(i * dataWidth) + j] = waterfallPallet[(int)(pixel * (WATERFALL_RESOLUTION - 1))];
+                    waterfallFb[(i * dataWidth) + j] = waterfallPallet[std::clamp<int>((int)(pixel * (WATERFALL_RESOLUTION - 1)), 0, WATERFALL_RESOLUTION)];
```
AlexandreRouma, line 509: No need to clamp a second time, the value is already clamped on line 508.

core/src/gui/widgets/waterfall.cpp
```diff
@@ -779,7 +777,7 @@ namespace ImGui {
             float dataRange = waterfallMax - waterfallMin;
             for (int j = 0; j < dataWidth; j++) {
                 pixel = (std::clamp<float>(latestFFT[j], waterfallMin, waterfallMax) - waterfallMin) / dataRange;
-                int id = (int)(pixel * (WATERFALL_RESOLUTION - 1));
+                int id = std::clamp<int>((int)(pixel * (WATERFALL_RESOLUTION - 1)), 0, WATERFALL_RESOLUTION);
```
AlexandreRouma, line 780: Value is already clamped on line 779.

core/src/signal_path/dsp.cpp
```diff
@@ -4,6 +4,10 @@ SignalPath::SignalPath() {
     
 }
 
+SignalPath::~SignalPath() {
+    stopFFT();
```
AlexandreRouma, line 8: This is not needed, the destructor of each member takes care of termination.

core/src/gui/widgets/waterfall.cpp
kpd200081: There was strange crash, like `pixel` gets somehow wrong value, may be fixed in master, I will check

rx888_source/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 cmake_minimum_required(VERSION 3.13)
 project(rx888_source)
 
+if (MSVC)
+  set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS TRUE)
```
AlexandreRouma, line 5: This is not needed for modules and is not recommended at all in general. Only use for the core software since it's needed for modules to bind to it.

rx888_source/CMakeLists.txt
```diff
@@ -9,14 +13,19 @@ else ()
...
-
 file(GLOB SRC "src/*.cpp")
-file(GLOB CYAPI "src/CyAPI/*.cpp")
 
 add_library(rx888_source SHARED ${SRC} ${CYAPI})
 target_link_libraries(rx888_source PRIVATE sdrpp_core)
 set_target_properties(rx888_source PROPERTIES PREFIX "")
 
+if (MSVC)
+  target_link_libraries(rx888_source PUBLIC Setupapi.lib)
```
AlexandreRouma, line 23: Why is this needed for a hardware module?

core/src/gui/widgets/combo_as_slider.cpp
```diff
@@ -0,0 +1,30 @@
+#include <gui/widgets/combo_as_slider.h>
```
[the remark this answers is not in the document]
kpd200081: I wrote this for usage in rx888_source - for selecting attenuation mostly (it has small step so many values)

core/src/signal_path/dsp.cpp
kpd200081: On master - yes, I will remove this

rx888_source/src/main.cpp
```diff
@@ -1,46 +1,127 @@
...
+
 class RX888SourceModule : public ModuleManager::Instance {
 public:
     RX888SourceModule(std::string name) {
         this->name = name;
 
-        if (!openFX3()) {
-            spdlog::error("No RX888 found!");
+        core::configManager.aquire();
+        std::string image = std::string(core::configManager.conf["resourcesDirectory"])+"/firmware/SDDC_FX3.img";
```
AlexandreRouma, line 49: Need to check that the firmware image exists.

rx888_source/CMakeLists.txt
kpd200081: Oh, I copy-paste this from somewhere

rx888_source/src/main.cpp
```diff
@@ -52,10 +133,12 @@ class RX888SourceModule : public ModuleManager::Instance {
         handler.stream = &stream;
         sigpath::sourceManager.registerSource("RX888", &handler);
 
-        spdlog::info("RX888SourceModule '{0}': Instance created!", name);
+        spdlog::info("RX888SourceModule '{0}'-'{1}'-v{2}: Instance created!", name, sddc_get_hw_model_name(sddc), sddc_get_firmware(sddc));
```
AlexandreRouma, line 136: Will cause a segfault if a null value is returned, needs checking

rx888_source/src/main.cpp
```diff
@@ -76,188 +159,261 @@ class RX888SourceModule : public ModuleManager::Instance {
...
     }
 
-    static void _worker(RX888SourceModule* _this) {
+#ifdef LIBSDDC_RAW
+    static void asyncHandler(uint32_t len, uint8_t *buf, void *ctx) {
+        RX888SourceModule* _this = (RX888SourceModule*)ctx;
         dsp::complex_t* iqbuffer = new dsp::complex_t[STREAM_BUFFER_SIZE];
+        int blockSize = len/sizeof(float);
+        for (int i = 0; i < blockSize; i++) {
+            iqbuffer[i].re = (float)buf[i*sizeof(float)] / 32768.0f;
```
AlexandreRouma, line 342: Use volk for type conversion instead of loop

rx888_source/src/main.cpp
```diff
@@ -76,188 +159,261 @@ class RX888SourceModule : public ModuleManager::Instance {
...
-                }
-            }
+        // Apply translation
+        volk_32fc_s32fc_x2_rotator_32fc((lv_32fc_t*)iqbuffer2, (lv_32fc_t*)iqbuffer, phaseDelta, &phase, blockSize);
 
-            // Write to output stream
-            _this->stream.swap(blockSize);
+        // Decimate
+        for (int d = 0; d < _this->deciminator+1; d++) {
+            blockSize = (blockSize >> 1);
```
AlexandreRouma, line 357: This decimation is incorrect. Data has to be filtered to the lowest bandwidth before.

rx888_source/src/main.cpp
```diff
@@ -76,188 +159,261 @@ class RX888SourceModule : public ModuleManager::Instance {
...
 
         // Deallocate buffers
         delete[] iqbuffer;
     }
+#else
+    static void asyncHandler(uint32_t len, float *buf, void *ctx) {
+        RX888SourceModule* _this = (RX888SourceModule*)ctx;
+        unsigned N = len / 2;   // I,Q,I,Q....
+        for (int i = 0; i < N; i++) {
+            _this->stream.writeBuf[i].re = (float)buf[i * 2];
```
AlexandreRouma, line 380: Like said above, just send the raw int16 using a stream and do type convertion later

AlexandreRouma:
Hello,

Please review the comments I've made to your pull request.

Best regards,
Alex

core/src/gui/widgets/combo_as_slider.cpp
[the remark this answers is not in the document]
AlexandreRouma: Imo a stepper slider widget would be better for that applciation.

CMakeLists.txt
kpd200081: With libusb but without libsddc it self this will not compile, maybe switch off by default and wait until libsddc will be more finished?

core/src/gui/widgets/combo_as_slider.cpp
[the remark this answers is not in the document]
kpd200081: This is stepped slider, but with custom label.

The main idea was that this will be combobox (selecting some values) but in form of slider (because of the occupied space)

core/src/gui/widgets/combo_as_slider.cpp
kpd200081: I rewrote this part, please take a look.

core/src/gui/widgets/waterfall.cpp
kpd200081: I removed clamp here.

rx888_source/src/main.cpp
kpd200081: I added checking, but it isn't really needed/

rx888_source/src/main.cpp
kpd200081: Fixed

rx888_source/src/main.cpp
kpd200081: I removed this part of code, now it use decimation from libsddc.

core/src/gui/widgets/waterfall.cpp
kpd200081: I rewrote this place and added comment to explain why this is needed.

kpd200081:
Thanks Alex. I fixed some parts  of code and reply on other comments, please take a look.

Best regards,
Pavel.

core/src/gui/widgets/waterfall.cpp
AlexandreRouma: I can't accept those changes. The values here should NEVER be NAN. The only case it could be NAN is if your source module is giving bad data. Checking for NAN here decrease performance.

AlexandreRouma:
Hello,
I've commented a few more things. I think it would be better to revert all the changes you made to code that's not for the libsddc support from this pull request.

I might take over the libsddc stuff tomorrow as I'll have a bit of free time and can implement it to support all devices.

Regards,
Alex

core/src/gui/widgets/waterfall.cpp
kpd200081: Ok, I understood. I was thinking that this can be useful if source produce bad data by some reason (which happened for me sometimes, so now I will try to figure why such data produced and try to solve this in source).

kpd200081:
Hello.

Ok, I will revert changes in waterfall, can I leave custom slider or should come up with something else?

All devices supported by libsddc should be already supported by this code because I was trying to write it dor dynamic checking of parameters supported by hardware. If you will found something locked to RX888 please let me know. Also I will rename module to indicate that it support not only RX888 but also all other radios from this family.

Best regards,
Pavel.

cozycactus:
can you help me to build it on macOS?
I have error during compilation

`[ 62%] Building C object source_modules/sddc_source/CMakeFiles/sddc_source.dir/src/libsddc/Core/arch/linux/ezusb.c.o
error: invalid argument '-std=c++17' not allowed with 'C'
make[2]: *** [source_modules/sddc_source/CMakeFiles/sddc_source.dir/src/libsddc/Core/arch/linux/ezusb.c.o] Error 1
make[1]: *** [source_modules/sddc_source/CMakeFiles/sddc_source.dir/all] Error 2
make: *** [all] Error 2
`

PR #114, opened 2021-04-26
Standardized sorting of bands array for band allocation plans

JoshuaKimsey:
* Standardized organisation of the bands array in the band allocation JSON files. They are now organised by the starting frequency.

* Added 'Polar Orbiting Satellites' to the general JSON file, as that affects all general users.


JoshuaKimsey:
I have added guides to the Contribution file for how to properly style and organise the JSON files for Band allocation identifiers and colour maps. Let me know what you think of the additions I made and if you like how I worded it.

PR #116, opened 2021-04-27
Add Canadian Bandplan

Starman0620:


root/res/bandplans/canada.json
```diff
@@ -0,0 +1,207 @@
...
+    {
+      "name": "TV Channels 2-4",
+      "type": "broadcast",
+      "start": 54000000,
+      "end": 72000000
+    },
+    {
+      "name": "FM Broadcast",
+      "type": "broadcast",
+      "start": 76000000,
```
JoshuaKimsey, line 143: Though I am not familiar with Canadian radio bands, is this correct? I'm not sure I've ever seen FM commercial radio frequencies this low?

JoshuaKimsey (commented):
Everything looks good! Just check on that FM commercial radio frequency and make sure it's correct.

root/res/bandplans/canada.json
Starman0620: Honestly I've lived here my whole life and can't recall hearing of any stations at such low frequencies. However, it is clearly how the band is supposed to be according to the government websites I ripped the values from.

Sources:
https://www.ic.gc.ca/eic/site/smt-gst.nsf/vwapj/2014_Canadian_Radio_Spectrum_Chart.pdf/$file/2014_Canadian_Radio_Spectrum_Chart.pdf
https://www.ic.gc.ca/eic/site/smt-gst.nsf/eng/sf10759.html#t2

JoshuaKimsey: You are right, and apparently other countries go down to 76MHz as well, including Japan. I honestly don't know any normal radio that goes down that low, and here in the US that's an area we broadcast TV channels on. Maybe it's used for more specialty applications, like digital radio or something? Oh well, works for me! In that case, everything seems good!

JoshuaKimsey (approved):

Issue #107, opened 2021-04-24
More de-emphasis options needed


JoshuaKimsey:
Firstly, let me say how amazing SDR++ is! I easily see this being able to compete with the likes of GQRX and Cubic on Linux. I can't wait for plugin support to be added so some SDR# plugins can be used! (Meteor demod would be so amazing!)

Now, for the issue I found. Normally in CubicSDR, when using the Stereo FM, I set de-emphasis to 25, as it provides a good middle ground between being a bit muffled and too tinny/noisy. Currently, the settings for SDR++ only has none, 50, and 75, which kinda presents a bit of an issue. If you could add in some more options, or the ability to set a custom value, either by hand or with a slider, it'd really help with regards to audio quality on the WFM, and even more so when stereo FM is added.

Thanks for the amazing work you're doing and I can't wait to see how SDR++ develops! I'm a full-time Linux user, so if you need any help testing out stuff for Linux, I'd be glad to offer my assistance! :)

AlexandreRouma:
Hello,

> I can't wait for plugin support to be added so some SDR# plugins can be used!

There is already plugin support. In fact, everything from each device to the radio is its own plugin. It's however impossible to add compatibility with SDR# for several reasons:

- Most SDR# plugins expect a windows environment
- SDR# plugins are written in C# which is extremely hard to interface from C++ (especially on linux/osx)
- SDR# plugins require the SDR# core libs which are not under any kind of open source license

> (Meteor demod would be so amazing!)

There is already a meteor demodulator plugin. It will soon be merged with the Weather Sat live decoder plugin that can decode down to image anything from LRPT (meteor) to to HRPT for NOAA and Meteor.

> Currently, the settings for SDR++ only has none, 50, and 75

There are international standard for preemphasis on WFM broadcast (75 for US and South Korea, 50 for the rest of the world). No other value is used. If you want a more muffled sound, you can sent the output sample rate to something like 12500Hz and you'll get a similar affect to the deemphasis filter. In the future there will also be audio noise reduction in the radio module. I think adding a custom option for the deemphasis would likely confuse users.

Best regards,
Alex

JoshuaKimsey:
> There is already plugin support. In fact, everything from each device to the radio is its own plugin.

Ah! Ok, that's also really awesome and should, in theory, make SDR++ very extensible for additions in the future. Like I said, I see it easily competing, and maybe beating, GQRX and CubicSDR on Linux. 

[...]

PR #118, opened 2021-04-27
Added more de-emphasis options

JoshuaKimsey:
* Added options for 10µs, 25µs, and 32µs de-emphasis options, to bring it in-line with what CubicSDR offers. This gives users more options for how much or little de-emphasis they desire. Done in reference to issue #107.

* Moved the 'none' option to the top of the drop-down menu from the bottom, as this makes more sense to most normal users. (At least in the USA, maybe it's different in other parts of the world?)

### DIFF
--- a/radio/src/wfm_demod.h
+++ b/radio/src/wfm_demod.h
@@ -179,7 +179,7 @@ public:
     }
 
     void setDeempIndex(int id) {
-        if (id >= 2 || id < 0) {
+        if (id >= 5 || id < 0) {
             deemp.bypass = true;
             return;
         }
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
Hello,

Thanks for adding those, but as I said in issue https://github.com/AlexandreRouma/SDRPlusPlus/issues/107,
I don't think adding more values to deemphasis is a good idea. I'm not quite sure why cubic does it since it doesn't make much sense.
Additionally, the way it's implemented means the default is no deemphasis at all and users with existing configs will get No to 36us deemphasis. This means it's very likely that they'll think the audio quality has either worsened or that something's broken.

Most people don't actually know broadcast standards or what that setting does at all which is why I set it to 50us (which sounds pretty "right" in both america and the rest of the world) and only allow real values.

Regards,
Alex

JoshuaKimsey:
Hey, that's fine. I just added them cause I usually liked using the 25us option best. But if you just wanna stick to what is standard then that's fine with me as well. I might still suggest moving the 'none' option to the top of the list though. It's a bit confusing to see it at the bottom. If that makes it be set as the default option, then it shouldn't be too hard to get a more hard-coded default set, then let the user change it to their preferences. Perhaps provide a value for the id initially to set it to 50 us at initial startup?

I'll look and see if I can do that. Like I said before, C++ is not my strongest area of knowledge, but my Java training means I can work with smaller stuff at least! :)

PR #120, opened 2021-04-28
Moved 'none' option in de-emphasis menu to top, Improved initial default de-emphasis selection

JoshuaKimsey:
* Moved the 'none' option for de-emphasis to the top of the drop-down menu to be more user friendly.

* Set default de-emphasis option for initial start to 50µs, without issues involving menu order. Achieved by improving if-statement selection parameters for determining bypassing de-emphasis.

(Tested multiple times for consistency of results. This should work exactly the same as it did, just with 'none' at the top of the menu. Default de-emphasis selection (50us) should be identical to before.)


JoshuaKimsey:
* Added request for capital letters to be used for hex colour codes in colour map JSON files. This is for consistencies sake.

* Fixed minor spelling error that was found in Contributing.md

JoshuaKimsey:
Hey @AlexandreRouma, what are your thoughts on this? Like the changes I made or are their other changes you want me to make?

AlexandreRouma:
Hello,
First of al, sorry didn't notice your pull request in my emails last time.

So, my issue with this modification is it's functionally equivalent, but will cause current users upgrading to have their current de-emphasis setting change from 50 to none or 75 to 50. Both of these will result in incorrect sound. As I said last time, most users don't know what the setting does, won't notice it has change and will just think the audio quality is worst than other software or earlier versions.

Plus, since support for other de-emphasis values isn't planned, it's a modification that isn't really useful. I chose to put the none setting at the end so it doesn't seem like a default (because it should never be used on broadcast).

I think the best cours of action if you really want other de-emphasis value is a custom fork of the radio module to do it.

Regards,
Alex

PR #160, opened 2021-07-06
Improve documentation for windows build

mbiette:
Hello,

I just started having a look at the code of SDR++ and, as I was going through the setup of my environment on Windows using Visual Studio 2019, I made a few changes:
- I updated the documentation to add details for the Visual Studio setup.
- Also, as I didn't find a binary release for RtAudio I simply installed it using vcpkg and adapted the build to take that into account. I think it simplify the setup so I added it in this pull request.

Please, let me know what you think

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 build/
 .vscode/
+.vs/
+.idea/
 *.old
 *.dll
 *.exe
@@ -9,3 +11,5 @@ build/
 root_dev/
 sdrpp_v0.2.5_beta_x64
 sdrpp_v0.2.5_beta_x64_new_wf
+Folder.DotSettings.user
+CMakeSettings.json
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,5 @@ build/
 *.wav
 .DS_Store
 root_dev/
-sdrpp_v0.2.5_beta_x64
-sdrpp_v0.2.5_beta_x64_new_wf
 Folder.DotSettings.user
 CMakeSettings.json
--- a/readme.md
+++ b/readme.md
@@ -52,7 +52,7 @@ The prefered IDE is [VS Code](https://code.visualstudio.com/) in order to have s
 
 ## Install dependencies
 * [cmake](https://cmake.org)
-* [vcpkg](https://vcpkg.io) (Below we assume it is installed in `C:\src\` as suggested in the documentation.)
+* [vcpkg](https://vcpkg.io)
 * [PothosSDR](https://github.com/pothosware/PothosSDR) (This will install libraries for most SDRs)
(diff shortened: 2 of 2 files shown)


mbiette:
Regarding the RtAudio change, there were a few steps in the build process that I missed but that are now fixed -- hence the few force-push on my branch. The final build using the workflow looks good now.

AlexandreRouma:
Hello, thanks for contributing but I see several porblems here:

1) SDR++ Isn't meant to be compiled using Visual Studio. It should be developped using only CMake and VSC (this way developing on all platform is the same). Adding support for VS adds uncessery bloat and potention bad practices created by VS.

2) RTAudio is meant to be installed on its own rather than from vcpkg, last time I checked the version on vcpkg was broken and I had to compile and install from source instead (hence why it's not used with vcpkg

3) the do_always target is meant to copy only the core DLLs when running for development. the audio_sink module (only one using RtAudio) isn't part of the SDR++ core.

I think it would be better to keep the clarity inhancements to the reamde (like adding links to vcpkg, pothos, etc...) but rolling back anything that has to do with VS or the audio sink.

mbiette:
Thank you for your feedback!

1. I understand concerning VS and will remove that. I will change the documentation to reflect the fact that the preferred IDE is VS Code.

2. RtAudio is installed the same way by vcpkg than by the hard coded steps (i.e. it downloads from git and build it as I can see in the PR that added it https://github.com/microsoft/vcpkg/pull/14774/files). As per my testing this works perfectly. So it is up to you if you want to keep it or not. Here the advantage is that it is one less build to manage and it gets integrated in CMake properly.

3. I see what you mean about `do_always`. Although from the code it looked like it was actually a step to copy all the DLLs that each library depends on, for running in dev as well as for the final build in the workflow. Adding this step and removing the step in the make_windows_package.ps1 is simply a way to ensure that either in dev or in the workflow the library is there. This is obviously due to the fact that I used vcpkg to install rtaudio but I still see it as an improvement to not have to build and install rtaudio separately. 

Do you confirm you do not want any of the change regarding the audio_sink?

Best regards,
Maxime

AlexandreRouma:
First of all sorry for the time to reply I got busy with something else this afternoon.

I've tried the RtAudio version from vcpkg again and I still have the crashing issues I was having last time for some reason, I'll have to investigate it later but for now I think keeping the current RtAudio situation is better.

For the do_always target, I prefer the current way of doing it. I want to make sure that, if needed, the core can be compiled completely separately for any modules. Thus keeping the do_always target to only copy core libraries is better imo.

mbiette:
I cleaned up the pull request and left only the .gitignore and readme.md changes. Let me know if that's good with you.

AlexandreRouma:
Alright perfect, merging.
Thanks for the contribution.

PR #164, opened 2021-07-11
Fix destructor crash due to wild pointer

mbiette:
Otherwise when RingBuffer is deleted before its init() being called it would crash in the destructor due to its _buffer pointer not being initialized.

### DIFF
--- a/core/src/dsp/buffer.h
+++ b/core/src/dsp/buffer.h
@@ -8,11 +8,7 @@ namespace dsp {
     template <class T>
     class RingBuffer {
     public:
-        RingBuffer():
-            _buffer{nullptr}
-        {
-
-        }
+        RingBuffer() {}
 
         RingBuffer(int maxLatency) { init(maxLatency); }
 
@@ -202,7 +198,7 @@ namespace dsp {
         }
 
     private:
-        T* _buffer;
+        T* _buffer = NULL;
         int size;
         int readc;
         int writec;
--- a/core/src/dsp/buffer.h
+++ b/core/src/dsp/buffer.h
@@ -8,7 +8,9 @@ namespace dsp {
     template <class T>
     class RingBuffer {
     public:
-        RingBuffer() {
+        RingBuffer():
+            _buffer{nullptr}
+        {
 
         }
 


AlexandreRouma:
Hello,
Although this could technically caused a crash if the program were to terminate before the RingBuffer::init() function is called, this in pratice cannot happen (declaring a RingBuffer without initializing it is just useless code).

This is a good code quality improvement though so I guess it's fine.
However:

I'd prefer that the way it's implemented match the code style of the rest of the software. I use NULL (which since C++11 is the same as nullptr) everywhere else in the code (imo it's cleaner and more noticable).
I also prefer initializing members directly in the class rather than before the constructor.

I'll make these changes and merge, please keep those in mind for future PRs. Will be sure to add that to the contributing.md as it seems to not be included.

Thanks for contributing.

Regards,
Alex

mbiette:
The value of NULL is actually still [implementation defined](https://en.cppreference.com/w/cpp/types/NULL). I'm using MSVC to compile on Windows and it has `#define NULL 0`. Another example, in the book "Effective Modern C++" by Scott Meyers which talks about C++11/14, he advises to prefer nullptr rather than NULL or 0 for pointers as it is the only way to make sure it is never interpreted as an integer.

Indeed, it is an improvement as it avoids issues related to the constructor without argument that is explicitly provided, There are a few others like that that I ran into and that I will push soon.

In the module I'm working on I did an initialization a bit later than I should it seems and part of the code that create instances of the sinks seems to create/delete/create the modules which made me run into this issue.

See src/signal_path/sink.cpp:240 
```
void SinkManager::loadStreamConfig(std::string name) {
   ...
    if (stream->running) {
        stream->sink->stop();
    }
    delete stream->sink;
    SinkManager::SinkProvider prov = providers[provName];
    stream->providerId = std::distance(providerNames.begin(), std::find(providerNames.begin(), providerNames.end(), provName));
    stream->sink = prov.create(stream, name, prov.ctx);
    if (stream->running) {
        stream->sink->start();
    }
    ...
}
```

AlexandreRouma:
yes that is normal. Sink class are created when needed and destroyed when switching sinks (and often at loading because by default the Null sink is selected)

PR #165, opened 2021-07-11
Fix more wild pointers

mbiette:


AlexandreRouma:
I think something cleaner than checks everywhere should be implemented. In my opinion the cleanest way would be to have a `bool _block_init = false;` private member and an `assert(_block_init);` in the very beginning of every **non worker** function.

I'll get to work implementing this.

PR #166, opened 2021-07-12
Fix reserved keywords

mbiette:
https://en.cppreference.com/w/cpp/keyword

"identifier that begins with an underscore followed by an uppercase letter is always reserved"

### DIFF
--- a/airspy_source/src/main.cpp
+++ b/airspy_source/src/main.cpp
@@ -585,7 +585,7 @@ private:
     std::string sampleRateListTxt;
 };
 
-MOD_EXPORT void _INIT_() {
+MOD_EXPORT void SDRPP_MOD_INIT() {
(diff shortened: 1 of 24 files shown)


AlexandreRouma:
This isn't issue as those are exported C symbols, they are not mangled and thus cannot interfere with any C++ keyword or symbol.
In addition pushing this change would break all current binary releases of plugins or partially updated plugins.
I don't think I can accept this PR.

mbiette:
These are reserved identifiers by the compilers in any contexts for [C](https://en.cppreference.com/w/c/language/identifier) or [C++](https://en.cppreference.com/w/cpp/language/identifiers). Using them obviously currently works but is undefined behavior none the less.
If it were to be merged it should probably be done as part of a new major version indeed.

PR #167, opened 2021-07-12
Added error log to the HackRF start and stop calls

mbiette:
The libhackrf fails to close the device properly which makes the entire lib crash when trying to start it again after that.
When the device is not closed correctly, we exit/init the libhackrf again to make sure it is in a state that will allow starting again later on. This probably will require more investigation on the root cause...

### DIFF
--- a/hackrf_source/src/main.cpp
+++ b/hackrf_source/src/main.cpp
@@ -228,8 +228,8 @@ private:
             return;
         }
 
-        auto err = static_cast<hackrf_error>(hackrf_open_by_serial(_this->selectedSerial.c_str(), &_this->openDev));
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
I cannot replicate this crash on any of my devices. is this related to SoapySDR?
If yes I'd rather not add a dirty work around like this which 100% causes a memory leak.

mbiette:
I will investigate more later where it is coming from. Here is the error message:
`[error] Could not close HackRF 0000000000000000457863c82256a31f: Input/Output Error`
I can update the pull request to only keep the logs if you want.

mbiette:
I changed the commit to only keep the additional logs in the hackrf module.

PR #175, opened 2021-07-16
Audio sink based on PortAudio

mbiette:


joshuarubin:
Why was this closed and not merged?

AlexandreRouma:
The code with portaudio had way more issues on linux than the current code so I decided to keep the current sink for the time being.

joshuarubin:
Ah, ok, well it would be really helpful if there was a way to force pulse over jack either at compile or runtime. I'm using pipewire, so it defaults to jack and was awful. Forcing it to pulse fixed things for me.

PR #190, opened 2021-07-27
Further progress the rigctl server towards full feature parity

thedocruby:
I found myself struggling to use the included rigctl server with many other programs that work fine with the HamLib rigctl server. After digging i realized this was mostly because the included rigctl server was unfinished and nowhere near feature parity with HamLib's server. This PR is my attempt to remedy that.

### DIFF
--- a/rigctl_server/CMakeLists.txt
+++ b/rigctl_server/CMakeLists.txt
@@ -14,6 +14,7 @@ file(GLOB SRC "src/*.cpp")
 include_directories("src/")
 include_directories("../recorder/src")
 include_directories("../meteor_demodulator/src")
+include_directories("../radio/src")
 
 add_library(rigctl_server SHARED ${SRC})
 target_link_libraries(rigctl_server PRIVATE sdrpp_core)
--- a/rigctl_server/src/main.cpp
+++ b/rigctl_server/src/main.cpp
@@ -405,11 +405,31 @@ private:
                 client->write(strlen(buf), (uint8_t*)buf);
             }
             else if (parts[0] == "set_mode") {
+                std::lock_guard lck(vfoMtx);
+
+                // if number of arguments isn't correct, return error
+                if (parts.size() != 3) {
+                    spdlog::error("Rigctl client sent invalid command: '{0}'", cmd);
+                    resp = "RPRT 1\n";
+                    client->write(resp.size(), (uint8_t*)resp.c_str());
+                    return;
+                }
+
+                // If not controlling the VFO, return
+                if (!tuningEnabled) {
+                    resp = "RPRT 0\n";
+                    client->write(resp.size(), (uint8_t*)resp.c_str());
+                    return;
+                }
+                
+                // If client is querying, respond accordingly
                 if(parts[1] == "?") {
                     resp = "FM WFM AM DSB USB CW LSB RAW";
                     client->write(resp.size(), (uint8_t*)resp.c_str());
                     return;
                 }
+                
+                // Parse mode and bandwidth
                 int mode;
                 switch(parts[1]) {
                     case "FM" : mode = RADIO_IFACE_MODE_NFM;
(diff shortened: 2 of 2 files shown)


AlexandreRouma:
Hello,

I think this was of parsing commands is a bit hacky. i'll have a look again at the spec tomorrow to handle long commands and compound commands. The manual didn't seem to mention those.

thedocruby:
I know the long commands are on the manual because `man rigctld` is how I'm getting most of my information, and I saw the backslash method mentioned multiple times. However, I didn't see the compound commands on the manual either, and they didn't even work on my official (although obsolete) build of rigctl that I got from apt. I had to clone the latest source from GitHub myself and `make install`. Then the compound commands worked, indicating that they are probably a more recent feature. The only reason I knew it should have worked is that my third-party program was constantly polling the SDR++ rigctl server with the command `fmv`, presumably to update the program's values for frequency, mode/bandwidth, and selected vfo. Of course, sdrpp threw an error every time and nothing was ever updated. So I set out to help fix that. I am still working on the PR, and have started work on commands for getting and setting the mode, bandwidth, vfo, antenna, and snap interval. I am absolutely open to feedback and review whenever you feel like providing it, but I would prefer if you waited until the PR is ready for review. I should be finished with the next commit sometime tomorrow, and then it won't be long before the PR is ready for review.

By the way, I hope I don't seem pushy or controlling. I am truly impressed by this masterpiece that you have diligently and delicately assembled, and it is definitely my favorite SDR software. The level of dedication you bring to this project daily is astounding, and it makes me self-conscious of my own endeavors as a developer, which seem flimsy and insignificant in comparison. To be honest, I am thrilled to contribute to such carefully crafted software, and I just cant wait to see the first official release. Thanks for the opportunity!

thedocruby:
I decided not to do antenna commands just yet, since that would be different for different sources. Probably should rework the command handling a bit before I do that.

AlexandreRouma:
Hello,

As I said, the way you implemented the changes is not up to the standards of the codebase both in style and in quality:

1) You have two parts of the code to execute the exact same command when you can just do two comparisons in one if statement and not have the code be twice is long and complicated for no reason.
2) You didn't respect the brace style for many of the ifs.

AlexandreRouma:
I fixed all the issues mentioned above, please pull and add commands to the new code if you want to do so.

thedocruby:
> 1. You have two parts of the code to execute the exact same command when you can just do two comparisons in one if statement and not have the code be twice is long and complicated for no reason.
This is true, and it was a pain to manage both at once, but I didn't see a way to combine them and retain compound command functionality. That being said, I didn't really try to figure out a good way to do that, so i guess that was just me being lazy.

> 2. You didn't respect the brace style for many for `if` statements. (see contributing.md)
Sorry, that's my mistake. I should have read the rules over before assuming that i could skip whitespace to save space on those long if-else chains.

> 3. You're not checking whether the VFO belongs to a radio before calling the interface with `RADIO_IFACE_CMD_SET_MODE`.
I guess I just assumed that the only module capable of creating a VFO was a radio, which is a dumb assumption to make. My bad.

> 4. Calling the vfo manager's `setBandwidth` function is **not** the right way to update the radio's frequency... Instead this has to be done either by calling the radio's interface using `RADIO_IFACE_CMD_SET_BANDWIDTH` or by manipulating the vfo widget through the waterfall.
I did consider using RADIO_IFACE_CMD_SET_BANDWIDTH instead of doing it through the VFO manager, but ultimately decided not to because it seemed to do the same thing, and I was trying to match the method used for getting the bandwidth that was already there.

All in all, this code was rather clumsy and messy, and I should have done better. I'm sorry for causing you extra work.

thedocruby:
I resolved the conflicts, but you might want to double check it, some of them were rather confusing.

AlexandreRouma:
Fixed some invalid changes (LOS and AOS are important commands for use with gpredict)

PR #288, opened 2021-08-20
Fix rigctl server to work with WSJT-X

3cky:
Currently, SDR++ can't be controlled from WSJT-X software due to some rigctl protocol handling discrepancies. Also, some rigctl commands required by WSJT-X are not implemented by SDR++. This PR fixes the SDR++ rigctl server to work with WSJT-X.

### DIFF
--- a/rigctl_server/src/main.cpp
+++ b/rigctl_server/src/main.cpp
@@ -427,20 +427,22 @@ private:
                 return;
             }
 
-            // if number of arguments isn't correct or the VFO is not "VFO", return error
+            // if number of arguments isn't correct, return error
             if (parts.size() != 3) {
                 resp = "RPRT 1\n";
                 client->write(resp.size(), (uint8_t*)resp.c_str());
                 return;
             }
 
-            // Check that the bandwidth is an integer
+            // Check that the bandwidth is an integer (0 or -1 for default bandwidth)
+            int pos = 0;
             for (char c : parts[2]) {
-                if (!std::isdigit(c)) {
+                if (!std::isdigit(c) && !(c == '-' && !pos)) {
                     resp = "RPRT 1\n";
                     client->write(resp.size(), (uint8_t*)resp.c_str());
                     return;
                 }
+                pos++;
             }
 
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
Hello,

Looks good however those changes are invalid. The newlines are added later in the code.

image

3cky:
Hello Alexandre!

The main difference is that SDR++ reports mode and bandwidth value as a single space-delimited line, like `FM 15000`. But WSJT-X expects response as _two_ lines, one for mode and the second for bandwidth value. [Gqrx rigctl server](https://github.com/csete/gqrx/blob/272bd75e8cb33ff54d2d8f71295295dcdbf4b962/src/applications/gqrx/remote_control.cpp#L551) also reports `get_mode` command result as a two-line string.

To illustrate, I connected the rigctl client to a dummy rigctld server, both from Hamlib package:

```
$ rigctld -m 1 -T 127.0.0.1 -t 4532 # startup dummy rigctl server

$ rigctl -m 2 -r 127.0.0.1:4532 # rigctl client
rigctl.c(158) Startup: rigctl -m 2 -r 127.0.0.1:4532

Rig command: M
Mode: LSB
Passband: 3000

Rig command: m
Mode: LSB
Passband: 3000

```

Please see below the raw port 4532 traffic sniffed with wireshark:

```
M LSB 3000
RPRT 0
m
LSB
3000
```

Hope this clarifies the situation.

AlexandreRouma:
Strange, the rigctl manual says that returned info has to be on a single line and space seperated

AlexandreRouma:
I'll switch back to your version. Hopefully this doesn't cause issues with https://github.com/AlexandreRouma/SDRPlusPlus/pull/190

3cky:
Before creating this PR, I checked gpredict and it works with patched SDR++ with no issues.

MMSSolutions:
When will this fix be incorporated in to the windows build? V1.0.3 doesn't seem to work with WSJT and I need it quickly :-)

PR #300, opened 2021-08-23
Fix Wrong Source Branch within BladeRF build

NickEngmann:
within the bladeRF build, if it reaches the Darwin branch it attempts to add /usr/local/include into the airspyhf_source, instead of the bladerf_source.

This PR fixes what I expect was a typo.


AlexandreRouma:
Thanks for the fix!
Looks like I copy pasted and forgot to change it out

PR #328, opened 2021-08-29
soapy_source: always show the Refresh button

ericek111:
When SDR++ starts with no devices found by Soapy, there is no Refresh button and you need to restart the app to be able to select any. This makes the Refresh button always visible.

After connecting a device and clicking the button:


AlexandreRouma:
Alright checks seem to have passed.

By the way, I see you're using soapysdr for the SDRplay, Don't do that, SoapySDRPlay is horribly broken.

SDR++ has native support for SDRplay devices, you just need to install the SDRplay API (it can't be shipped with SDR++ for licensing reasons)

PR #329, opened 2021-08-29
soapy_source: refresh devices on start if none in devList

ericek111:
If there are no devices in the device list enumerated when the application starts, clicking the play button causes the whole app to crash (probably because of nullptr in devList). This tries to ask Soapy for devices before giving up gracefully, providing smoother experience to the user and preventing the crash.

Clicking the play button, however, still makes SDR++ think it's running. There is no feedback between source modules and the main app. Some interface should be made to return status data to the app for proper state keeping.

### DIFF
--- a/soapy_source/src/main.cpp
+++ b/soapy_source/src/main.cpp
@@ -296,6 +296,14 @@ private:
     static void start(void* ctx) {
         SoapyModule* _this = (SoapyModule*)ctx;
         if (_this->running) { return; }
+        if (_this->devId < 0) {
+            _this->refresh();
+            _this->selectDevice(config.conf["device"]);
+            if (_this->devId < 0) {
+                return;
+            }
+        }
+
         _this->dev = SoapySDR::Device::make(_this->devArgs);
 
         _this->dev->setSampleRate(SOAPY_SDR_RX, _this->channelId, _this->sampleRate);
--- a/soapy_source/src/main.cpp
+++ b/soapy_source/src/main.cpp
@@ -297,11 +297,7 @@ private:
         SoapyModule* _this = (SoapyModule*)ctx;
         if (_this->running) { return; }
         if (_this->devId < 0) {
-            _this->refresh();
-            _this->selectDevice(config.conf["device"]);
-            if (_this->devId < 0) {
-                return;
-            }
+            return;
         }
 
         _this->dev = SoapySDR::Device::make(_this->devArgs);
--- a/soapy_source/src/main.cpp
+++ b/soapy_source/src/main.cpp
@@ -297,6 +297,7 @@ private:
         SoapyModule* _this = (SoapyModule*)ctx;
         if (_this->running) { return; }
         if (_this->devId < 0) {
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
Hello,

I think it would be better to do nothing if no device is found to prevent the crash instead of quickly refreshing and playing.
This could lead to the user opening a device they didn't want to open and cause other crashes this time inside SoapySDR itself.

I'll add some feedback mechanism for modules to report if they've started properly (likely just returning true or false)

ericek111:
If the user wants control over which device gets selected, they have the option to exercise their caution by refreshing the list and selecting it manually (see #328).

Most users won't be connecting more than one SDR to their computer. This saves two clicks in case the dongle gets disconnected, the SDRplay service crashes for the thousandth time... I think it brings more good than evil.

AlexandreRouma:
> If the user wants control over which device gets selected, they have the option to exercise their caution by refreshing the list and selecting it manually

Users should not need to have caution when clicking play, the software has to be in the state visible to the user.

> Most users won't be connecting more than one SDR to their computer
Actually, a lot do. Especially ham radio operators like me with an always plugged in SDR used as a panadapter for my rig.

> the SDRplay service crashes for the thousandth time

I think this issue is beacause you're using SoapySDR for your SDRplay (at least I think from your other PR).
All other source modules instantly pick back up the device in case of a disconnect.

I'll push a slight change to avoid refreshing before the start and merge

ericek111:
> Users should not need to have caution when clicking play, the software has to be in the state visible to the user.

The user has picked "Soapy" as their source, clearly expressing their intention to use it. There are no devices available. So they connect one and can either click play (to start the "Soapy" source), or they always have the option to refresh the list manually.

> Actually, a lot do. Especially ham radio operators like me with an always plugged in SDR used as a panadapter for my rig.

A lot do, most don't. If you already have a device connected, it will be enumerated at the start of the app and will show in the list (so clicking "Play" will start the selected device). And even IF, by the slim chance, you started SDR++, THEN plugged in multiple SDRs, then clicked Play, it will just use the device that's specified in the config (at least by theory).

Yes, you already told me that Soapy is unstable on Twitter I think. It uses SDRPlay directly, so it's not any more stable or unstable, because it's always the libsdrplay API that crashes (and brings down SDR++ with it).

AlexandreRouma:
> , it will just use the device that's specified in the config (at least by theory).
It does not behave like that.
It'll select the first device it finds because it's the only device available. It's a fallback of the select function.

ericek111:
Yeah, I mean when you plug in multiple devices at the same time. It first enumerates, then selects using the value in config -- and falls back to the first device if it's not available. BUT, if your SDR saved in the config is amongst the newly connected devices, it will be pre-selected either after you click the Refresh button or Play. 

I understand this change is rather opinionated, that's why I made it into a separate PR. You have the final word, naturally, so if you feel like this brings more harm than good, so be it. I'm trying to convince you as a user who has already enjoyed this feature.

PR #330, opened 2021-08-30
Make the zoom slider exponential

ericek111:
Zooming on signals with small bandwidth on a scope of 10 MHz is a pain with linear slider. If the selected bandwidth is made to grow exponentially, it provides finer control towards the lower end of the range while maintaining enough granularity in the upper range.

This makes zooming, one of the most frequently touched control in the whole program, much more useful. Instead of having to grip my mouse and try to work inside 5 pixels to get a good look at a signal, I can now comfortably zoom in and out.

### DIFF
--- a/core/src/gui/main_window.cpp
+++ b/core/src/gui/main_window.cpp
@@ -598,8 +598,14 @@ void MainWindow::draw() {
     ImGui::SetCursorPosX((ImGui::GetWindowSize().x / 2.0) - (ImGui::CalcTextSize("Zoom").x / 2.0));
     ImGui::Text("Zoom");
     ImGui::SetCursorPosX((ImGui::GetWindowSize().x / 2.0) - 10);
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
Ok, float imprecisions completely break this.
In order, this shows `bw, normBw, factor, finalBw` at no zoom, you can see the bandwidth if pretty far off.
I'll remake that part of the code completely
image

AlexandreRouma:
Alright, fixed.

ericek111:
That did occur to me, but I didn't think the imprecision would affect the result so drastically. I suppose it does make sense if we're only using the fractional part.

You may want to remove this, though:
https://github.com/AlexandreRouma/SDRPlusPlus/blob/012903fbf478d32a0beeec93f9b9b81ca041dd40/core/src/gui/main_window.cpp#L608

AlexandreRouma:
yup just fixed that.

Normally the bandwidth should now be correct for any zoom level and sample rate

PR #352, opened 2021-09-11
Make Windows includes lowercase for MinGW

mnhauke:
MinGW-w64 ships all Windows SDK headers as lowercase, which prevents
cross-compiling this code from Linux.

Windows filesystems are case insensitive so it should work fine with
lowercase includes.


AlexandreRouma:
SDR++ should NOT be compiled with MinGW. MinGW has horrible performance compared to MSVC.

Issue #334, opened 2021-09-01
Mac build missing SDR Play


ve3ds:
Likely this is being fixed, but the MacOS version is missing the SDRPlay  drivers ... will they be there in 1.04? Great program and REALLY needed on MacOS ... really really needed. Thx DS

AlexandreRouma:
Unless SDRplay adds a way to install the library unattended, I can't include it in the prebuilt mac version.
Anyone who wishes to use SDR++ with an SDRplay SDR on mac currently has to build it themselves following the linux instructions.

ve3ds:
Alex Ok, I guess I’ll give it a try!! 
I’ve loaded all the dependencies to get the pre build Mac to run.  What do i need for SDRPlay? Soapy, or the proprietary drivers?
Thanks
73 DS

AlexandreRouma:
You need to install the API/drivers from SDRplay's website, then you should be able to use the following commands to build and install SDR++.

First, install the dependencies;
```sh
brew update
brew install fftw glew glfw volk airspy airspyhf portaudio hackrf rtl-sdr libbladerf
```

Then download the SDR++ source code:
```sh
git clone https://github.com/AlexandreRouma/SDRPlusPlus
```

Prepare the build:
```sh
cd SDRPlusPlus
mkdir build
cd build
cmake .. -DOPT_BUILD_SDRPLAY_SOURCE=ON -DOPT_BUILD_PLUTOSDR_SOURCE=OFF -DOPT_BUILD_SOAPY_SOURCE=OFF -DOPT_BUILD_BLADERF_SOURCE=OFF -DOPT_BUILD_AUDIO_SINK=OFF -DOPT_BUILD_PORTAUDIO_SINK=ON -DOPT_BUILD_NEW_PORTAUDIO_SINK=ON
```

Build and install:
```sh
make
sudo make install
```

ve3ds:
Ok… looks like I missed a few dependencies after all !

Thanks I’ll give it a go.

Thanks 
Dana VE3DS
in Toronto.

ve3ds:
I keep getting an error …

[...]

Issue #244, opened 2021-08-12
Doesn't seem to be working on Big Sur 11.5.1.


afpena:
It does install, and I can see it listed on usr/local/bin/, but whenever you try to run it... 

XXXXXX@yyyyyy ~ % sdrpp     
dyld: Library not loaded: /usr/local/opt/glew/lib/libGLEW.2.2.dylib
  Referenced from: /usr/local/bin/sdrpp
  Reason: image not found
zsh: abort      sdrpp

Sudoing doesn't do the trick either.

PiTiLeZarD:
Yup, same problem here on Big Sur (M1)

loewal:
Worked fine for me under Big Sur and even now under Monterrey.
But I must say that I used homebrew to install glew.

PiTiLeZarD:
Aha yeah so:

```
brew install glew
sudo mkdir -p /usr/local/opt/glew/lib/
sudo ln -s /opt/homebrew//lib/libGLEW.2.2.dylib /usr/local/opt/glew/lib/libGLEW.2.2.dylib
```

would get you there. But on the M1 I now have 

```
dyld: Library not loaded: /usr/local/opt/glew/lib/libGLEW.2.2.dylib
  Referenced from: /usr/local/bin/sdrpp
  Reason: no suitable image found.  Did find:
	/usr/local/opt/glew/lib/libGLEW.2.2.dylib: mach-o, but wrong architecture
	/opt/homebrew/Cellar/glew/2.2.0_1/lib/libGLEW.2.2.0.dylib: mach-o, but wrong architecture
[1]    84238 abort      sdrpp
```

Which meansI have to suss out how to install glew with brew using rosetta!

PiTiLeZarD:
Might be worth adding this to the doc? I don't know who maintains this, I can help if needed (for the doc, I'm not a C/C++ guy unfortunately ;))

xologram:
can confirm worked for me with "ibrew" method and parallel x86 homebrew!

afpena:
Worked after going through the entire glew installation. Thank you all for the support!

AlexandreRouma:
Likely fixed in the nightly but still no way to test. reopen if not.

Issue #234, opened 2021-08-08
Documentation updates:  Mac, FreeBSD build-from-source


pcf000:
I built sdr++ successfully from source on my Big Sur Mac -- I'm listening to local FM radio right now.  I swear it's better quality sound than other rtl-sdr projects I've tried.

I followed the Linux/BSD build instructions and used brew for the package dependences.  Some of them have different names, thus

    brew install librtlsdr soapysdr soapyrtlsdr cmake fftw glfw glew volk rt-audio

I needed to disable airspy, airspyhf, hackrf, and plutosdr modules.

I also built sdr++ successfully on FreeBSD on Raspberry Pi 4, but haven't managed to run it yet;  openGL doesn't seem to like a VNC connection.  Again, some dependences have different names, thus

    pkg install cmake fftw3 glfw glew volk SoapySDR SoapyRTLSDR libiio libad9361-iio rtaudio

I also installed mesa-dri and libosmesa while trying to get past the openGL errors, but I don't know if those are strictly necessary.

FWIW, both these systems use clang, which emits a lot of compiler warnings while building.

joe-azzara:
I had brew install librtlsdr soapysdr soapyrtlsdr cmake fftw glfw glew volk rt-audio
on mac os Big Sur, it went through but nothing else happens, then went to installed sdr++ from the releases page still nothing happened can anyone help???

pcf000:
Assuming that the window at least comes up for you, make sure to set your source and sink -- on mine, they're RTL-SDR and Built-In Output, and at least the source defaulted to something like File instead.

joe-azzara:
I had gone on discord and found out the problem. Thanks

aneitzel:
can you write a litle guide here whats needed to get it run under MacOS? Pleeeeeeeeeese..... like cloning from git to local drive and then?

MCKLtech:
@pcf000 , how did you "disable airspy, airspyhf, hackrf, and plutosdr modules."?

I'm getting an error of dyld: Symbol not found: _airspy_list_devices which I assume is related to this.

AlexandreRouma:
this error happens you have have an outdated version of libairspy or libairspyhf.

To disable modules, you can use the option names listed in the module list in the readme. For example, the airspy_source module uses the OPT_BUILD_AIRSPY_SOURCE parameter. To disable it, add the following argument to the cmake command: `-DOPT_BUILD_AIRSPY_SOURCE=OFF`

[...]

Issue #217, opened 2021-08-03
didn't work on Mac


cn0xroot:
```
sdrpp
dyld: Library not loaded: /usr/local/opt/glfw/lib/libglfw.3.dylib
  Referenced from: /usr/local/bin/sdrpp
  Reason: image not found
[1]    70951 abort      sdrpp
```

install via https://github.com/AlexandreRouma/SDRPlusPlus/releases/download/1.0.0/sdrpp_macos_amd64.pkg

osx 11.4

cn0xroot:
already install glfw with homebrew

aneitzel:
same here... installed glfw via macports before but SDRpp dont find lib at start. A step by step guide how to get it run under MacOS will be good.

AlexandreRouma:
Issue seems to be macport isntalling packages where the dynamic linker can't find them. Either configure macos's dynamic linker to find the packages (I have no idea how or if it's even possible, I don't use osx myself) or just use homebrew.

m1ctk:
It installed ok but I cant find it anywhere !

xologram:
same, installed but nowhere to be found. on m1 mac but figured should work via rosetta?

AlexandreRouma:
it does not create an icon, run it via the terminal by running `sdrpp`.

m1ctk:
> it does not create an icon, run it via the terminal by running `sdrpp`.

Doesn't work with Big Sur 11.5.1

xologram:
thanks @AlexandreRouma ! 

I too get the same error as OP:

`$ sdrpp
dyld: Library not loaded: /usr/local/opt/glew/lib/libGLEW.2.2.dylib
  Referenced from: /usr/local/bin/sdrpp
  Reason: image not found
Abort trap: 6`

since homebrow installs libGLEW in different location had to symlink the lib:

`mkdir -p /usr/local/opt/glew/lib/
sudo ln -s /opt/homebrew/lib/libGLEW.2.2.dylib /usr/local/opt/glew/lib/libGLEW.2.2.dylib` 

but now when I try to start it I get wrong arch :( 

`$ sdrpp
dyld: Library not loaded: /usr/local/opt/glew/lib/libGLEW.2.2.dylib
  Referenced from: /usr/local/bin/sdrpp
  Reason: no suitable image found.  Did find:
        /usr/local/opt/glew/lib/libGLEW.2.2.dylib: mach-o, but wrong architecture
[...]

PR #402, opened 2021-10-02
Document how to build on MacOS

corvus-ch:
I was able to build SDR++ on macOS Big Sur quite easily. So I thought I contribute with a bit of documentation. This might help with all those issues on that topic (#334, #234, #217, #244, #189).

To be honets, I have not tested this on a clean system. So I might have missed a dependency or two.

As I was not able to find a Homebrew package providing `libiio`, my solution was to disable PlutoSDR.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -50,7 +50,38 @@ There are currently no existing packages for other distributions, for these syst
 
 ## MacOS
 
-TODO
+You might get lucky by using the installer downloadable from [the Releases page](https://github.com/AlexandreRouma/SDRPlusPlus/releases).
+
+### Building it yourself
+
+This guide requires you to have Homebrew installed. Check out https://brew.sh/ on how to do so.
+
+Get yourself a copy of the source code. This can be done by downloading the
+source package from [the Releases
+page](https://github.com/AlexandreRouma/SDRPlusPlus/releases). Eextract it to a
+location of your choice. Alternativley you can clone
+https://github.com/AlexandreRouma/SDRPlusPlus.git using git.
+
+```sh
+brew install \
+  airspy \
+  airspyhf \
+  cmake \
+  glew \
+  glfw \
+  hackrf \
+  librtlsdr \
+  portaudio \
+  rt-audio \
+  soapyrtlsdr \
+  volk
+mkdir build
+cd build
+cmake .. -DOPT_BUILD_PLUTOSDR_SOURCE=false
+make -j$(sysctl -n hw.ncpu)
+```
+
+The above was tested on macOS Big Sur (11.5).
 
 ## BSD
 


K0HYD:
Corvus-ch!

Thanks for the jump start. I'm so close I can taste it. But no cigar! Managed to get the Make files created and compile the program. It runs but then crashes on launch with a missing symbol in a dyld. I've looked around but don't no where to go from here.

The pertinent error messages in the output are:

[2021-10-05 00:18:45.203] [info] Ready.
dyld[30324]: missing symbol called
—— and

"exception" : {"codes":"0x0000000000000000, 0x0000000000000000","rawCodes":[0,0],"type":"EXC_CRASH","signal":"SIGABRT"},
  "termination" : {"code":4,"flags":518,"namespace":"DYLD","indicator":"Symbol missing","details":["(terminated at launch; ignore backtrace)"],"reasons":["missing symbol called"]},
  "asi" : {"dyld":["missing symbol called"]},
——
 I did have a number of warning while running cmake. They were all the same and looked like this:

/Users/dalepuckett/SDRPP/SDRPlusPlus-master/core/src/dsp/resampling.h:115:13: warning: 'calcOutSize' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
        int calcOutSize(int in) {
            ^
-----

The only errors reported in the launch up to the crash were the missing modules I don't use anyway:

bladerf_source, limesdr_source, plutosdr_source, sdrplay_source and soapy_source

------

any help from anyone would be appreciated by this newbee!

73 de Dale - K0HYD

AlexandreRouma:
That warning isn't important, it's just complaining about some C++ synthax.
I've never seen that error but I've recently changed some CMake stuff so idk.
I'd need someone who has a working build to try to build the newer version and see if it works.

AlexandreRouma:
So, I think this means it's something specific to @K0HYD's system.

@K0HYD could you give more info about your mac? (mainly which MacOS version you're running)
Updating the libraries with brew could also be a good idea:
```
brew update
```
And then run the brew command you used to install the dependency again.

K0HYD:
Thank you for looking at this issue. No errors reported during compile. Crash occurs on launch after the readout on the console prints “Ready”

System here is 2017 iMac, i7. OS is MacOS 12 Beta 8 (latest).

Complete machine description is at the end of the attached CrashReport. The report says the problem is a missing symbol in dyld 4. The problem is I have no idea how to find what is missing from this information.

Hopefully, the attached crash report will help someone more familiar with the process to zero in on the issue.

I appreciate the help! I would really like to be able to do these builds. I’m so close it’s frustrating.

I’ve also attached the Terminal output during the make and the launch. I though it might be helpful.

Thanks again!

73 de Dale - K0HYD



On October 5, 2021 at 9:46:26 AM, AlexandreRouma ***@***.***) wrote:

So, I think this means it's something specific to @K0HYD's system.

@K0HYD could you give more info about your mac? (mainly which MacOS version you're running)
Updating the libraries with brew could also be a good idea:

brew update
And then run the brew command you used to install the dependency again.

corvus-ch:
Somtimes my brain needs a bit longer. I have now checked the github action workflow for how it is done there. I have update the example in the readme accordingly.

Jakketet:
"Old Guy" trying to compile SDR++ on an "Old Mac", MBP on High Sierra but stuck with following 
"make" stops with 
```
/Users/jt/SDRPlusPlus/core/src/config.cpp:5:10: fatal error: 'filesystem' file not found
#include <filesystem>
```
someone a clue?
suppose some missing library...

73's de ON1TJ

PR #417, opened 2021-10-08
avoid division by zero as long as waterfall height is 0

thomasgi1:


AlexandreRouma:
That's not possible, onResize() will ALWAYS be called on the first framed rendered when the UI shows up.
If it's called any time before there's something wrong elsewhere in the code.

PR #419, opened 2021-10-08
renamed SpyServer source module class

thomasgi1:
renamed SpyServer source module class AirspyHFSourceModule -> SpyServerSourceModule, so that it does not collide with the original Airspy HF source module causing massive memory corruption


thomasgi1:
I think it depends on the compiler/linker how it handles the two identical sets of symbols AirSpyHFSourceModule*. I’m compiling on MacOS 11.6 using Apple clang version 13.0.0. I guess you precompiled version used another compiler.

AlexandreRouma:
The precompiled version is build on macOS 10.15 also with clang

Issue #145, opened 2021-05-26
macos audio sink not worked


BG7JAF:
Hello all:
I build the application from source and it can run up. But there is only some noise out. I look at the log message, seems that  it's something wrong with the audio buffer. Here is the log file.
By the way, recorded file is OK.

OS: macos 11.2.3
version: 0.3.0_beta

[log.txt](https://github.com/AlexandreRouma/SDRPlusPlus/files/6543454/log.txt)

porkiedev:
Bump!

I'm having this issue on Ubuntu MATE 20.04 LTS. Though I haven't built from source, I was considering trying it.
My audio sink sounds good sometimes, then it'll randomly start buffering like crazy, probably 8-10 buffers a second, so the audio is still intelligible but really annoying to listen to since it keeps freezing.

AlexandreRouma:
This is a known issue. Until I can get my hands on a real mac I can't fix this.

> I'm having this issue on Ubuntu MATE 20.04 LTS
there is no link to the OSX issue.
This issue is likely RtAudio trying to use Alsa instead of pulseaudio. You need to build RtAudio from source and disable the Alsa backend.

PR #421, opened 2021-10-08
fixes #248 and #145: increased schedule priority for rtaudio engine

### LINKED ISSUE #248: Audio cut out MacOS 11.4
Very grateful to have this app working natively on the Mac.  Beautiful audio when SDRPP is the frontmost window, but audio cuts out momentarily and intermittently when transitioning to another app or window or when other operations unrelated to SDRPP take resources.  It settles back to smooth sound pretty quickly.

thomasgi1:


thomasgi1:
I found the reason, why (new) port audio sink did not work. It was not used although it was configured to do so. Instead still the rtaudio sink was used. The reason for that was the same as with the SpyServer - AirspyHF module problem. All three type of audio sinks use the same names for their symbols. So they got mixed up, when loading them at the same time into memory. I renamed the symbols in port audio respectively in new port audio. Now I can use them and they work better than the original rtaudio sink. A new PR is on the way.

PR #422, opened 2021-10-09
renamed (new) port audio sink classes

thomasgi1:
renamed (new) port audio sink classes, so that the symbol name do not get mixed up with rtaudio sink, when loaded into memory at the same time

### DIFF
--- a/sink_modules/new_portaudio_sink/src/main.cpp
+++ b/sink_modules/new_portaudio_sink/src/main.cpp
@@ -26,7 +26,7 @@ SDRPP_MOD_INFO {
 
 ConfigManager config;
 
-class AudioSink : SinkManager::Sink {
+class NewPortAudioSink : SinkManager::Sink {
 public:
     struct AudioDevice_t {
         const PaDeviceInfo* deviceInfo;
@@ -38,7 +38,7 @@ public:
         std::string sampleRatesTxt;
     };
     
-    AudioSink(SinkManager::Stream* stream, std::string streamName) {
+    NewPortAudioSink(SinkManager::Stream* stream, std::string streamName) {
         _stream = stream;
         _streamName = streamName;
 
@@ -65,7 +65,7 @@ public:
         selectDevByName(selected);
     }
 
-    ~AudioSink() {
+    ~NewPortAudioSink() {
         stop();
         gui::mainWindow.onPlayStateChange.unbindHandler(&playStateHandler);
     }
@@ -171,7 +171,7 @@ public:
 
 private:
     static void playStateChangeHandler(bool newState, void* ctx) {
-        AudioSink* _this = (AudioSink*)ctx;
+        NewPortAudioSink* _this = (NewPortAudioSink*)ctx;
 
         // Wake up reader to send nulls instead of data in preparation for shutoff
         if (newState) {
@@ -329,7 +329,7 @@ private:
 
     static int _mono_cb(const void *input, void *output, unsigned long frameCount,
         const PaStreamCallbackTimeInfo* timeInfo, PaStreamCallbackFlags statusFlags, void *userData) {
-        AudioSink* _this = (AudioSink*)userData;
+        NewPortAudioSink* _this = (NewPortAudioSink*)userData;
         
         // For OSX, mute audio when not playing
         if (!gui::mainWindow.isPlaying()) { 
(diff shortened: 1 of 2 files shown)


thomasgi1:
I do not understand your ignorance. In fact the current implementation does not work on mac. Your precompiled version comes with both the port audio and new port audio module installed, which results in a roulette for the equally named symbols, when loading them into memory. So, I see three options to fix the problem:

1. give all the classes in the modules a unique name (rejected by you)
2. write a big documentation for the users and potential developers which a statement, that only one module having shared names must be present in the modules directory (I guess nobody will read that)
3. throw away all modules having reused symbol names NOW

I’m so sorry that I wasted my spare time in order to improve your software.

AlexandreRouma:
Do you not realize that by design all modules have identical symbols?
Literally all modules have a `_INFO_`, `_INIT_`, `_CREATE_INSTANCE_`, `_DELETE_INSTANCE_` and `_END_` symbol.... And THAT doesn't cause an issue???
Having identical symbol **doesn't matter** when **nothing links to them**.
As I said many times, you have an issue elsewhere triggering this "bug". And I'm willing to bet it's a misconfigured debugger.

thomasgi1:
I know that each module has those identical symbols. But they are referenced using dlsym respectively GetProcAddress giving the explicit shared library name:

#ifdef _WIN32
    mod.handle = LoadLibraryA(path.c_str());
    if (mod.handle == NULL) {
        spdlog::error("Couldn't load {0}. Error code: {1}", path, GetLastError());
        mod.handle = NULL;
        return mod;
    }
    mod.info = (ModuleInfo_t*)GetProcAddress(mod.handle, "_INFO_");
    mod.init = (void(*)())GetProcAddress(mod.handle, "_INIT_");
    mod.createInstance = (Instance*(*)(std::string))GetProcAddress(mod.handle, "_CREATE_INSTANCE_");
    mod.deleteInstance = (void(*)(Instance*))GetProcAddress(mod.handle, "_DELETE_INSTANCE_");
    mod.end = (void(*)())GetProcAddress(mod.handle, "_END_");
#else
    mod.handle = dlopen(path.c_str(), RTLD_LAZY);
    if (mod.handle == NULL) {
        spdlog::error("Couldn't load {0}.", path);
        mod.handle = NULL;
        return mod;
    }
    mod.info = (ModuleInfo_t*)dlsym(mod.handle, "_INFO_");
    mod.init = (void(*)())dlsym(mod.handle, "_INIT_");
    mod.createInstance = (Instance*(*)(std::string))dlsym(mod.handle, "_CREATE_INSTANCE_");
    mod.deleteInstance = (void(*)(Instance*))dlsym(mod.handle, "_DELETE_INSTANCE_");
    mod.end = (void(*)())dlsym(mod.handle, "_END_");
#endif

All other symbols are resolved by the dynamic linker. When it loads a shared library, it puts all the symbols in a process wide table. If there is a name collision with already existing symbols, then the linker has to decide, which of the competing symbols has to be left in the symbol table. At this point the different OS may have different decision strategies. I would not rely on such a strategy. Instead I would ensure that the symbol names are unique.

Reading the dlopen manuals on Linux and MacOS gives some interesting details:

Linux:

Zero or more of the following values may also be ORed in flag:
RTLD_GLOBAL
The symbols defined by this library will be made available for symbol resolution of subsequently loaded libraries.
RTLD_LOCAL
This is the converse of RTLD_GLOBAL, and the default if neither flag is specified. Symbols defined in this library are not made available to resolve references in subsequently loaded libraries.

MacOS:

One of the following flags may be ORed into the mode argument:

     RTLD_GLOBAL  Symbols exported from this image (dynamic library or bundle) will be available to any images build with -flat_namespace option
                  to ld(1) or to calls to dlsym() when using a special handle.

     RTLD_LOCAL   Symbols exported from this image (dynamic library or bundle) are generally hidden and only availble to dlsym() when directly
                  using the handle returned by this call to dlopen().

So, on Linux the default is RTLD_LOCAL, but on MacOS it is RTLD_GLOBAL! The default is applied here, because none of the flags is given in dlopen().

So, a forth option to fix the problem on MacOS is:

mod.handle = dlopen(path.c_str(), RTLD_LAZY | RTLD_LOCAL);

AlexandreRouma:
I updated the code to use RTLD_LOCAL. Does that fix the issue?

thomasgi1:
Yes, that solves the issue with mixed-up symbols causing memory corruption. I even temporary reverted my Spy Server module renaming in order to double check that no crash occurs, when using that module. It works fine now with RTLD_LOCAL.

However I’m still investigating another audio related bug (rtaudio and port audio), when using multiple VFOs. I will file an issue once I have more details.

AlexandreRouma:
To be honest I never got the RtAudio sink to work properly on MacOS, that's why it's not included in the prebuilt versions.

PR #490, opened 2021-11-01
Add file_source WAV formats 8-bit-unsigned and float32

### LINKED ISSUE #478: u8 IQ file playback
(SDR++ v1.0.4)
When I load a SDRSharp IQ recording (8 bit unsigned), the file plays with double speed.
[20180516_233225Z_405300000Hz_IQ.wav.gz](https://github.com/AlexandreRouma/SDRPlusPlus/files/7436128/20180516_233225Z_405300000Hz_IQ.wav.gz)
```
Channels       : 2
Sample Rate    : 250000
Precision      : 8-bit
Duration       : 00:00:10.00 = 2500000 samples ~ 750 CDDA sectors
File Size      : 5.00M
Bit Rate       : 4.00M
Sample Encoding: 8-bit Unsigned Integer PCM
```
```
00000000: 5249 4646 644b 4c00 5741 5645 666d 7420  RIFFdKL.WAVEfmt 
00000010: 1000 0000 0100 0200 90d0 0300 20a1 0700  ............ ...
00000020: 0200 0800 6461 7461 404b 4c00 7f80 7f80  ....data@KL.....
00000030: 7e7f 807f 8080 7f7f 807f 8080 8080 7f7f  ~...............
...
```
Another SDRSharp (8 bit unsigned) recording with sample rate 1024000 shows the same and the signal is all over the place.

However if I convert to 16 bit signed samples, everything is fine.

rs1729:
First step to support WAV format 8 bit unsigned (and float32) and to resolve #478.

Don't know how portable `struct __attribute__((__packed__))` is, however reading directly into the WavHeader-struct could also be problematic.

`bitDepth==8` is assumed to be 8 bit unsigned (e.g. SDRSharp IQ recordings).
`bitDepth==32` is assumed to be float32. Though gqrx float32 raw files that have a WAV header produced by sox have a non-PCM header.
The `float32Mode` checkbox would be obsolete.

A more flexible way reading the WAV header would be to find the "fmt " chunk to get the number of channels, sample rate and bits per sample, and then find the "data" chunk to determine the header size and beginning of the data samples. Then RF64 files (SDRConsole) could also be supported.

Not sure if it would be better to stop the playback when opening a new file, the worker thread has to be reset in any case.
What should happen if the file format is not supported?


AlexandreRouma:
File source requires  a rewrite anyayy so this will be added once it's done

PR #501, opened 2021-11-04
Use CMAKE_INSTALL_FULL_LIBDIR instead of hardcoding lib

drsn0w:
This prevents Gentoo's Portage from throwing strict-multilib errors as Gentoo only expects 32-bit libraries in lib and 64-bit libraries in lib64

This should not effect building on other distributions as cmake will set CMAKE_INSTALL_FULL_LIBDIR appropriately.


drsn0w:
This pull request was accidentally submitted prematurely. Closing for now until I can complete testing...

AlexandreRouma:
This is not needed, CMake adds the prefix automatically

shuyuan-liu:
As I understand it it's not about adding the prefix. The problem is some distros differentiate between lib and lib64 but SDR++ currently only installs to lib, not lib64.

drsn0w:
This is correct. I ran into some issues with the patch and will need to do
further testing on non-Linux OS’s.

PR #509, opened 2021-11-10
Tweak cmake logic to build on Fedora

lacombar:
Fedora is missing the glfw3 and volk packages. This PR allows to build SDR++ on Fedora, alongside with:

 # dnf install fftw-devel \
	glew-devel \
	SoapySDR-devel \
	libad9361-devel \
	libiio-devel \
	rtaudio-devel \
	hackrf-devel \
	python3-mako \
	airspyone_host-devel \
	rtl-sdr-devel

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,8 @@ option(OPT_BUILD_RIGCTL_SERVER "Rigctl backend for controlling SDR++ with softwa
 
 # Other options
 option(USE_INTERNAL_LIBCORRECT "Use an external version of libcorrect" ON)
+option(USE_INTERNAL_VOLK "Use an external version of volk" OFF)
+option(USE_INTERNAL_GLFW3 "Use an external version of glfw3" OFF)
 
 # Core of SDR++
 add_subdirectory("core")
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -1,4 +1,6 @@
 cmake_minimum_required(VERSION 3.13)
+include(ExternalProject)
+
 project(sdrpp_core)
 
 if (USE_INTERNAL_LIBCORRECT)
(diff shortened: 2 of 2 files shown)


AlexandreRouma:
Hello,

This is a bad idea:
1) A quick search reveals that both GLFW and Volk are available on fedora:
https://src.fedoraproject.org/rpms/volk
https://src.fedoraproject.org/rpms/glfw
2) Even if they weren't, they can simply be built from source
3) SDR++ is available on fedora too
https://src.fedoraproject.org/rpms/sdrpp
4) Including these libraries like this will mean that they will be considered part of SDR++ and break other software

lacombar:
1 & 3. Neither are integrated in Fedora 33 (not EOL'ed yet) or prior. Requiring a full system upgrade for a single software is not acceptable.
2. they can, but integration will have to be done manually
4. that argument would be valid if it was enabled by default, which is not the case here. Also, `libcorrect` is a precedent of a external library arbitrarily fully imported where `ExternalProject_Add` could have been used (I'll be happy to submit a patch for that).

AlexandreRouma:
> 1 & 3. Neither are integrated in Fedora 33 (not EOL'ed yet) or prior. Requiring a full system upgrade for a single software is not acceptable.
> 2. they can, but integration will have to be done manually

The URL I sent says GLFW is included in 33, it would be quite surprising that such a common library would be missing.
Also, building volk manually is the normal and recommended way or distros that have an outdate version or no version at all.

> 4. that argument would be valid if it was enabled by default, which is not the case here. Also, libcorrect is a precedent of a external library arbitrarily fully imported where ExternalProject_Add could have been used (I'll be happy to submit a patch for that if you want).

The issue isn't it being enabled or not, it's it being there in the first place. There must be a seperation between software and its dependencies for reasonable cases (and in this case a single library that has to be built manually because it's missing from the package manager on one specific version of a distros would be considered reasonable). Adding an option to download and build the source of some of its dependencies just doesn't make sense.
To stay consistent I'd have to also include these options for a bunch of other libraries that could be missing.

I'd recommend asking the fedora maintainers to make volk available for older releases if that's possible (it has been available in all major distros for over 5 years now so I don't understand how a version that's barely a year old doesn't have it)

Regards

lacombar:
> There must be a separation between software and its dependencies for reasonable cases
Counter-argument: software like `meshroom` (photogrametry software) includes everything needed in its build system.

> Also, building volk manually is the normal and recommended way on distros that have an outdated version or no version at all.
And my patch will help people stuck there, at no cost for the project (once again, this is *disabled* by default). On top of that it allows to test against arbitrary version of upstream library, which can help with issue in the future.

> To stay consistent I'd have to also include these options for a bunch of other libraries that could be missing.
I can help to work toward that. It would certainly help building SDR++ on BSD and environment which lack dependent libraries.

STBoyden:
> 1 & 3. Neither are integrated in Fedora 33 (not EOL'ed yet) or prior. Requiring a full system upgrade for a single software is not acceptable

You can install individual pieces of software on an older version of Fedora from a newer version's repository. For example: `dnf install --releasever=34 volk`. Doing so will update the version of GLIBC from 2.32 to 2.33, but should otherwise be completely fine to do.

lacombar:
> You can install individual pieces of software on an older version of Fedora from a newer version's repository. For example: `dnf install --releasever=34 volk`. Doing so will update the version of GLIBC from 2.32 to 2.33, but should otherwise be completely fine to do.

Upgrading the glibc is not a trivial thing to do in a critical system I can't afford to have any downtime.

That being said, the focus should not really be about Fedora per-se, but more generally, if someone want to use SDR++ on a platform missing the dependent library (non-mainstream CPU architecture / software environment), should they have to do all the work by themselves (and in such a way, most likely making them consider other SDR solutions).

Said otherwise, should SDR++ be portable to non-mainstream systems out-of-the-box ? Once again, I'll be happy to do the work.

STBoyden:
I think the issue is that it's not up to SDR++ to install its dependencies, unless in a very specific cases where the dependency is:

1. hard to get/find
2. needs to be built in a very specific environment

It seems that most, if not all, mainstream distros have volk and glfw. In the case where they don't have a version of volk available in their distro's package manager, then it is better for them to build it manually on their system rather than requiring SDR++ to build it for them. As volk is freely available and is relatively easy to build, then there is no need for SDR++ to build it alongside itself.

lacombar:
Short of discouraging future contributors, what is the negative impact of facilitating accessibility of SDR++, especially if 1) someone else is doing the job, 2) the changes are minimal, and 3) changes are not enabled by default ?

Unless of course, discouraging contributors is the goal.

Issue #529, opened 2021-12-01
Segfault on start with Debian Bullseye + AMD Radeon 21.40


aeburriel:
Upon starting, sdrpp 1.0.4 segfaults with the following backtrace:

> #0  __rawmemchr_avx2 () at ../sysdeps/x86_64/multiarch/memchr-avx2.S:65
> #1  0x00007ffff75ef5a2 in _IO_str_init_static_internal (sf=sf@entry=0x7fffffffd900, ptr=ptr@entry=0x0, size=size@entry=0, pstart=pstart@entry=0x0) at strops.c:41
> #2  0x00007ffff75c336f in _IO_strfile_read (string=0x0, sf=0x7fffffffd900) at ../libio/strfile.h:95
> #3  __GI___isoc99_sscanf (s=0x0, format=0x7ffff7b3abbc "%d.%d") at isoc99_sscanf.c:28
> #4  0x00007ffff7abc8d9 in ImGui_ImplOpenGL3_Init(char const*) () from /usr/lib/libsdrpp_core.so
> #5  0x00007ffff79e589c in sdrpp_main(int, char**) () from /usr/lib/libsdrpp_core.so
> #6  0x00007ffff7591d0a in __libc_start_main (main=0x55555556f410 <main>, argc=1, argv=0x7fffffffe1d8, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, 
>     stack_end=0x7fffffffe1c8) at ../csu/libc-start.c:308
> #7  0x000055555556f4fa in _start ()

Which points to ImGui ticket [#4279](https://github.com/ocornut/imgui/issues/4279)

Updating ImGui to 1.85 solves the issue on my system.

AlexandreRouma:
pretty sure that was a bug in imgui that was fixed

PR #530, opened 2021-12-01
Update Dear ImGui to 1.85

aeburriel:
Update ImGui to 1.85, solves ticket #529

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -56,9 +56,9 @@ if (MSVC)
     # Volk
     target_link_libraries(sdrpp_core PUBLIC volk)
 
-    # Glew
-    find_package(GLEW REQUIRED)
-    target_link_libraries(sdrpp_core PUBLIC GLEW::GLEW)
+    # GL
+    find_package(OpenGL REQUIRED)
+    target_link_libraries(sdrpp_core PUBLIC OpenGL::GL)
 
     # GLFW3
     find_package(glfw3 CONFIG REQUIRED)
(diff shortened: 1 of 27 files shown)


aeburriel:
Because in ImGui 1.85 OpenGL is initialized by the embedded loader:
https://github.com/ocornut/imgui/issues/4445

If you review the demos you'll see it's now the standard way.

AlexandreRouma:
Has this been properly tested yet?
I afraid of switching to this embedded loader without being sure it still handles OpenGL 2.0 (ES) properly on raspberry pi boards

aeburriel:
Nope, only in a couple of Linux/amd64.
Unfortunately I don't have access to other architectures.

AlexandreRouma:
Alright, I'm gonna wait for this branch to be tested on other platforms before merging.
But before, a few things:
1) If glew is not longer needed, it should be removed from the core CMakeLists.txt
2) It's important to check if this change has any performance impact.

aeburriel:
Code updated as follows:
1. Re-added an include required for linking properly in Windows which I removed by accident.
2. Removed all GLEW dependencies and updated readme.

Tested on Linux/amd64 & Windows/x64.
About the performance, I agree it should be tested in low-end devices. Dear ImGui update is mostly bugfixes, according to its changelog. Graphics access remains the same, the only difference is the way OpenGL is initialized.

AlexandreRouma:
Alright, i'll check that the ImGui update was done correctly, some settings had to be changed in imconfig for the last version

AlexandreRouma:
Looks like trying to get GL from pkg-config files on MacOS, not really sure what the cleanest way to go about it is at this point.

AlexandreRouma:
Alright, I've done this along with other changes in the pull request I just merge.
Please confirm that the latest commit fixes the issues you were having and re-open if it doesn't

PR #538, opened 2021-12-07
Add GOES Demodulator for testing / recording HRIT signals

racerxdl:
I copied the Meteor Demod plugin and adapted for GOES 16/17 signaling so people could use that for debugging issues with the RX. It doesnt do much, but it does save the demodulated I/Q same format as meteor. 

I'm thinking about making a sink that can output to SatHelperApp  and manage the ingestion system directly from SDRPP or even add  the ingestion system to SDRPP.

image

My current signal isnt great, but its decodable. The PSK demod and constellation view is working as expected.

Let me know if any changes required in the code ^^

btw: nice api, I really liked it :D

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ CMakeSettings.json
 poggers_decoder
 m17_decoder/libcorrect
 SDR++.app
+cmake-build-debug/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,7 @@ option(OPT_BUILD_METEOR_DEMODULATOR "Build the meteor demodulator module (no dep
 option(OPT_BUILD_RADIO "Main audio modulation decoder (AM, FM, SSB, etc...)" ON)
 option(OPT_BUILD_NEW_RADIO "Beta Main audio modulation decoder (AM, FM, SSB, etc...)" ON)
 option(OPT_BUILD_WEATHER_SAT_DECODER "Build the HRPT decoder module (no dependencies required)" OFF)
+option(OPT_BUILD_GOES_DEMODULATOR "Build the GOES demodulator module (no dependencies required)" OFF)
 
 # Misc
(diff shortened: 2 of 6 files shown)


AlexandreRouma:
Hi,

Looks good, though I'd prefer to wait to get back to my weather sat decoder module instead of having yet another standalone decoder, I'll keep the PR open until this code can be properly merged with other satellite decoders.

By the way, I'd recommend spending some time tuning the costas bandwidth M&M parameters since using the ones from the meteor demodulator might result in inferior decodes.

racerxdl:
Sure, let me know then! 

About the parameters, I used the ones from my SatHelperApp decoder. These seen to work fine for overall usage of GOES. GOES postprocessing is a bit overwhelming for the CPU, so I'm not sure if its expected SDR++ to be able to ingest stuff, or just put out through TCP/UDP the HRIT Frames. What do you think?

AlexandreRouma:
It should be possible.
The goal of the wheather sat decoder module will be to decode satellites all the way to the final data without needing external software, so it's expected to do the post processing. Of course that processing has to be done intelligently to save CPU.

racerxdl:
The heavy part from GOES is the image processing. People like false color, and for 25 megapixel images, thats cpu / memory intensive (thats why in raspbery pi they disable image processing). 

Still, might be just good to put a checkbox, then the user can decide what its enabled (like SatHelperApp does)

3Domse3:
Where can I find that plugin?

3Domse3:
> @3Domse3 only in this PR and you need to build it yourself.

Whats a PR? Is there a beginner-friendly guide for building on Windows by any chance?

racerxdl:
> > @3Domse3 only in this PR and you need to build it yourself.
> 
> Whats a PR? Is there a beginner-friendly guide for building on Windows by any chance?

PR = Pull request, here where you're posting the messages.

So far there is no begginer friendly tutorial on how to use it, and also only the demodulator, it doesnt output the images and other stuff that comes from GOES.

PR #582, opened 2022-01-12
[libcorrect] initialize bestpath to 0

dankamongmen:
I understand that you're vendoring a copy of libcorrect (for whom I'm also submitting this), but given that they haven't committed since 2018, I figured I'd post it here, too. Eliminates a compiler warning by initializing to a somewhat dubious 0. I feel there ought be an actual error sentinel here, but no such luck with the unsigned type; at least return a known something.


AlexandreRouma:
Hello,
If you look at the code this doesn't actually matter since `buf->num_states` is never zero. This means it'll always end up assigned to a certain value.
Haven't got any warnings about it so not sure why your compiler complains.
I prefer leaving it exactly as in the original code for now.

dankamongmen:
sounds good, thanks for the quick reply. agreed that it functionally doesn't matter.

PR #591, opened 2022-01-16
fix warnings about unused/maybe uninitialized variables

kistlin:
the removed variables are reported unused with gcc and the flag  "-Wall".
The warning level could be increased in general. To keep it clean and find possible bugs early.

### DIFF
--- a/core/src/config.cpp
+++ b/core/src/config.cpp
@@ -36,7 +36,7 @@ void ConfigManager::load(json def, bool lock) {
         file >> conf;
         file.close();
     }
-    catch (std::exception e) {
+    catch (const std::exception& e) {
         spdlog::error("Config file '{0}' is corrupted, resetting it", path);
         conf = def;
         save(false);
--- a/core/src/config.cpp
+++ b/core/src/config.cpp
@@ -36,7 +36,7 @@ void ConfigManager::load(json def, bool lock) {
         file >> conf;
         file.close();
     }
-    catch (const std::exception& e) {
+    catch (std::exception e) {
         spdlog::error("Config file '{0}' is corrupted, resetting it", path);
         conf = def;
         save(false);
--- a/core/src/core.cpp
+++ b/core/src/core.cpp
@@ -337,7 +337,7 @@ int sdrpp_main(int argc, char* argv[]) {
 #else
     const char* glsl_version = "#version 120";
     GLFWmonitor* monitor = NULL;
-    for (int i = 0; i < OPENGL_VERSION_COUNT; i++) {
+    for (size_t i = 0; i < OPENGL_VERSION_COUNT; i++) {
         glsl_version = OPENGL_VERSIONS_GLSL[i];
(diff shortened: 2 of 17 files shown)


core/src/config.cpp
```diff
@@ -36,7 +36,7 @@ void ConfigManager::load(json def, bool lock) {
         file >> conf;
         file.close();
     }
-    catch (std::exception e) {
+    catch (const std::exception& e) {
```
AlexandreRouma, line 39: Please revert, just a style string that barely does anyhing for performance, I don't care much about const correcteness

core/src/core.cpp
```diff
@@ -337,7 +337,7 @@ int sdrpp_main(int argc, char* argv[]) {
 #else
     const char* glsl_version = "#version 120";
     GLFWmonitor* monitor = NULL;
-    for (int i = 0; i < OPENGL_VERSION_COUNT; i++) {
+    for (size_t i = 0; i < OPENGL_VERSION_COUNT; i++) {
```
AlexandreRouma, line 340: please revert, that value  (manually entered, it's a) would never get passed 5 or 6 there's no reason to use size_t which just takes more space for no reason

core/src/core.cpp
```diff
@@ -450,8 +450,12 @@ int sdrpp_main(int argc, char* argv[]) {
     bandplan::loadColorTable(bandColors);
 
     bool _maximized = maximized;
-    int fsWidth, fsHeight, fsPosX, fsPosY;
-    int _winWidth, _winHeight;
+    int fsWidth = 0;
```
AlexandreRouma, line 453: please revert, these values are never used if not assigned (try to understand the rest of the code) and declaring them like this just waste space and is less clean

core/src/gui/main_window.cpp
```diff
@@ -489,12 +483,11 @@ void MainWindow::draw() {
         ImGui::SetColumnWidth(1, winSize.x - menuWidth - 60);
         ImGui::SetColumnWidth(2, 60);
         ImGui::BeginChild("Left Column");
-        float menuColumnWidth = ImGui::GetContentRegionAvailWidth();
 
         if (gui::menu.draw(firstMenuRender)) {
             core::configManager.acquire();
             json arr = json::array();
-            for (int i = 0; i < gui::menu.order.size(); i++) {
+            for (size_t i = 0; i < gui::menu.order.size(); i++) {
```
AlexandreRouma, line 490: Same issue as above, no reason to use a size_t here, if your menu is over 2 billion elements in length you have other issues

core/src/utils/networking.cpp
```diff
@@ -289,7 +288,7 @@ namespace net {
                 }
                 entry.handler(std::move(client), entry.ctx);
             }
-            catch (std::exception e) {
+            catch (const std::exception& e) {
```
AlexandreRouma, line 291: here too I don't care much about const correctness since thus won't affect performance and just make things less readable

AlexandreRouma:
Alright, tests seem to have passed, I'll merge it once I get the time to fix the number of merge issues it'll cause with all the changes I did to my local version.

PR #593, opened 2022-01-17
Update main.cpp

### LINKED ISSUE #573: SDR++ version 1.04 server issue connecting to WSJT-X version 2.5.3.
Have not been able to successfully connect to rig control using a RTL-SDR v3.  Hamlib library installed is w64 version 4-4.  Local host and port is 127.0.0.1:4532.  SDR++ Rigctl Server is running and initially shows "Connected" to WSJT-X.  WSJT-X reports connection error and after re-entering radio control parameters, reports hamlib error.  SDR++ Rigclt Server then reverts to "listening".
2022-01-01 09_23_32-Window
2022-01-01 09_23_04-Window
Same error occurs with or without firewall being active.

Thanks

giordyb:
Current Hamlib release ( >= 4.x) sends one extra parameter (the VFO) to the call. Also VFO alone isn't valid, it needs to be either one of VFOA/B/SUB/MAIN, it might fix #573

### DIFF
--- a/misc_modules/rigctl_server/src/main.cpp
+++ b/misc_modules/rigctl_server/src/main.cpp
@@ -379,7 +379,7 @@ private:
             std::lock_guard lck(vfoMtx);
 
             // if number of arguments isn't correct, return error
-            if (parts.size() != 2) {
+            if (parts.size() != 3) {
                 resp = "RPRT 1\n";
                 client->write(resp.size(), (uint8_t*)resp.c_str());
                 return;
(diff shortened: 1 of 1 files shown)


giordyb:
> Haven't looked into it much but I'm pretty sure this breaks it for every other hamlib version.

I've done some more research and it seems that the rigctld backend is written **without** the VFO mode support which is fine and correctly recognized and handled by the rigctl client, even on Hamlib 4.4, so I don't think the updated Hamlib version is the problem.

I submitted this PR because I am writing some python code that uses Hamlib and was having the same issue as  #573.
The proposed fix is a workaround but as you said will break older versions.
I now believe that Hamlib should automatically detect if it's talking to a non-vfo mode backend and deal with it accordingly but for some reason it's not, so I'm closing the PR.

PR #596, opened 2022-01-19
Add Fedora 35 RPM build

JoshuaMulliken:


AlexandreRouma:
hello,
I only make builds of distros I actually know how to support.

PR #610, opened 2022-01-23
add Hamlib 4.x partial compatibility

giordyb:
Hi,

Hamlib 4.x introduced a new "vfo_mode" to rigctld that changes the way commands are sent by default. rigctl handles this automatically but any software that uses the apis is required to send the "set_vfo_mode" command to the rigctld server to switch back to the "old" command modes (see https://github.com/Hamlib/Hamlib/issues/930).

the only change required in SDR++ is to accept the `set_vfo_mode(0)` command, this way the API knows that it needs to talk to the backend using the old command modes.

I was able to test this and control SDR++ via the Hamlib APIs without any issues by using this python code and the changes you see in this PR.

```
import sys

sys.path.append("/usr/local/lib/python3.9/site-packages/")
import Hamlib
import time

Hamlib.rig_set_debug(Hamlib.RIG_DEBUG_VERBOSE)

my_rig = Hamlib.Rig(Hamlib.RIG_MODEL_NETRIGCTL)
my_rig.set_conf("rig_pathname", "localhost:4533")
my_rig.set_conf("retry", "5")
my_rig.open()
my_rig.set_vfo_opt(0)

my_rig.set_freq(Hamlib.RIG_VFO_CURR, 145000000)
my_rig.set_mode(Hamlib.RIG_MODE_FM)

my_rig.set_freq(Hamlib.RIG_VFO_CURR, 123000000)
my_rig.set_mode(Hamlib.RIG_MODE_AM)
``` 

thanks,

Giordano


AlexandreRouma:
Hello,
I didn't know this was all that was needed to get hamlib 4.0 to talk to older apps.
I'll try it out before merging to check if it actually fixes the issues.

giordyb:
> Hello,
> I didn't know this was all that was needed to get hamlib 4.0 to talk to older apps.
> I'll try it out before merging to check if it actually fixes the issues.

HI @AlexandreRouma,

Hold on, I don't think it will fix the WSJTX issue since the code that will talk to SDR++ will need to both use the current master version of Hamlib AND add that command after opening the connection.

But if someone that might be developing something using the new Hamlib library needs to talk to apps that use the older protocol (like SDR++) now he/she can do it.

I think ideally it would be better to have the SDR++ rigctld have 2 sets of commands (one for the old version and one for the new) and have a configurable switch somewhere to pick between the two.

AlexandreRouma:
> I think ideally it would be better to have the SDR++ rigctld have 2 sets of commands (one for the old version and one for the new) and have a configurable switch somewhere to pick between the two.

This is quite hacky and adds lots of boilerplate. 

Honestly I think the hamlib developpers did a big mistake by making such a big change instead of doing the opposite and adding a command to set the protocol version. that way software could have checked if the server supports it and only use it if it is. Now we're faced with a huge amount of software that still use an old version of the rigctl protocol that are completely broken for people who updates and if these software update to the new hamlib protocol then it won't work for people still using the old version. Completely ridiculous.

giordyb:
@AlexandreRouma nevermind

PR #611, opened 2022-01-24
Debugflags now working out of the box in Visual Studio (Windows)

oelison:
Actual missing references to debug lib from Soapy and RtAudio. Actually manual adjusted, but working then.


AlexandreRouma:
hello,
The way compiler flags work in the cmake setup is going to change to avoid modifying 33 files you you did, I'll make it work when the change is done

PR #615, opened 2022-01-25
Satisfy GCC -Wformat-security with string literals

hewittc:
GCC 11.1 fails to build the application without this change, emitting "format not a string literal and no format arguments [-Wformat-security]".

### DIFF
--- a/core/src/gui/menus/module_manager.cpp
+++ b/core/src/gui/menus/module_manager.cpp
@@ -43,10 +43,10 @@ namespace module_manager_menu {
                 ImGui::TableNextRow();
 
                 ImGui::TableSetColumnIndex(0);
-                ImGui::Text("%s", name.c_str());
+                ImGui::TextUnformatted(name.c_str());
 
                 ImGui::TableSetColumnIndex(1);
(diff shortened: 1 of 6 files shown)


AlexandreRouma:
It's just a warning, not an error. Adding a format string like this just makes performance worse and the code uglier.
I suggest looking for the actual error that prevents compilation.

hewittc:
Seems to be a result of CFLAGS injected through default Arch Linux makepkg.conf. PKGBUILDs in AUR might be broken as a result.
```
CFLAGS="-march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions \
        -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security \
        -fstack-clash-protection -fcf-protection"
```

AlexandreRouma:
that is already a more reasonable change.
I'll merge it and make sure the rest of the code is consitent in using `ImGui::TextUnformatted`
I'd however suggest the arch maintainer remove this quite stupid blanket error.

I'll make sure this change doesn't have any unwanted graphical differences.

Issue #445, opened 2021-10-16
Sound card issues on Xubuntu 20.04 LTS


VE2ZAZ:
Greetings,

I have installed SDR++ v1.0.3 on a Xubuntu 20.04LTS desktop, been using it with a SDRPlay RSP1A, and wish to adopt the software. However I have been experiencing a series of sound card issues with SDR++, mostly it being unable to access the proper sound card on my system.

In the Sinks area, "Audio" does not even show up unless I run the software as root. All I get is "None". Here is what I get when I launch SDR++ as the default user in a console:
_[2021-10-16 12:13:26.388] [info] SDR++ v1.0.3
[2021-10-16 12:13:26.388] [info] Loading config
[2021-10-16 12:13:27.004] [info] Loading icons
[2021-10-16 12:13:27.018] [info] Loading band plans
[2021-10-16 12:13:27.035] [info] Loading band plans color table
[2021-10-16 12:13:27.055] [info] Loading modules
[2021-10-16 12:13:27.055] [info] Loading /usr/lib/sdrpp/plugins/limesdr_source.so
[2021-10-16 12:13:27.069] [info] Loading /usr/lib/sdrpp/plugins/weather_sat_decoder.so
[2021-10-16 12:13:27.086] [info] Loading /usr/lib/sdrpp/plugins/new_portaudio_sink.so
[2021-10-16 12:13:27.103] [info] Loading /usr/lib/sdrpp/plugins/plutosdr_source.so
[2021-10-16 12:13:27.120] [info] Loading /usr/lib/sdrpp/plugins/rtl_tcp_source.so
[2021-10-16 12:13:27.136] [info] Loading /usr/lib/sdrpp/plugins/sdrplay_source.so
[2021-10-16 12:13:27.153] [info] Loading /usr/lib/sdrpp/plugins/rigctl_server.so
[2021-10-16 12:13:27.169] [info] Loading /usr/lib/sdrpp/plugins/network_sink.so
[2021-10-16 12:13:27.187] [info] Loading /usr/lib/sdrpp/plugins/frequency_manager.so
[2021-10-16 12:13:27.203] [info] Loading /usr/lib/sdrpp/plugins/recorder.so
[2021-10-16 12:13:27.219] [info] Loading /usr/lib/sdrpp/plugins/file_source.so
[2021-10-16 12:13:27.236] [info] Loading /usr/lib/sdrpp/plugins/hackrf_source.so
[...]

Issue #105, opened 2021-04-22
No audio output, core dump


darrenfreeman:
I am using an SDRPlay as the source, under Ubuntu 20.10.

I have so far not heard any sound output from SDR++, and it hasn't appeared in my PulseAudio sound control as an application that's producing sound. After changing the audio sink UI control between the various options, I got a core dump:

```
[2021-04-22 20:29:42.611] [info] SDRPlaySourceModule 'SDRplay Source': Tune: 7247000.0!
[2021-04-22 20:30:14.107] [info] RtAudio stream open
[2021-04-22 20:30:25.008] [info] RtAudio stream open
[2021-04-22 20:30:27.846] [info] RtAudio stream open
[2021-04-22 20:31:02.257] [info] RtAudio stream open
[2021-04-22 20:31:10.238] [error] Could not open audio device
terminate called after throwing an instance of 'RtAudioError'
  what():  RtApi:: a stream is not open!
Aborted (core dumped)
```

Here is another example.

```
[2021-04-22 20:34:26.067] [info] SDRPlaySourceModule 'SDRplay Source': Tune: 0.0!
[2021-04-22 20:34:26.505] [error] Could not open audio device
[2021-04-22 20:34:26.525] [info] SDRPlaySourceModule 'SDRplay Source': Tune: 7247000.0!
[2021-04-22 20:34:26.525] [info] Ready.
[2021-04-22 20:34:26.525] [info] SDRPlaySourceModule 'SDRplay Source': Tune: 7247000.0!
terminate called after throwing an instance of 'RtAudioError'
  what():  RtApi:: a stream is not open!
Aborted (core dumped)
```

AlexandreRouma:
hello,

Seems like a bad install of RtAudio. Please uninstall your current version and reinstall RtAudio from source.

darrenfreeman:
Uninstalled the rtaudio packages. Compiled the git master and installed. From the output of cmake,

```
-- Checking for module 'rtaudio'
--   Found rtaudio, version 5.1.0
```

Now, if I try to change the audio output, or even if I simply set the sink from Audio to None,

```
[2021-04-22 22:45:02.176] [info] RTLSDRSourceModule 'RTL-SDR Source': Tune: 0.0!
[2021-04-22 22:45:02.583] [warning] ConfigManager locked, waiting...
[2021-04-22 22:45:02.631] [error] Could not open audio device
[...]

PR #679, opened 2022-03-05
Avoid an audio_sink Crash and Provide RtAudio Bug Workaround

SamRH:
I can see you're working on possible replacements for the current audio_sink so I haven't properly reworked the logic around doStart() and AudioSink::running. Happy to do that if you would prefer. This bug does result in SDR++ crashing on startup from the get go on one of my machines, and on the other once I select a device that won't be able to be opened SDR++ crashes every time until I manually adjust audio_sink_config.json

I've also documented an RtAudio bug that causes issues on a default fresh install of Fedora 35 and will likely cause issues on other Linux distributions (like RedHat 8+). I'm going to look at submitting a patch to RtAudio for that. Happy to split that out of this PR or move it to the wiki or somewhere else if you don't want it here.

Thanks for the tool! I've only just started playing around with SDR and SDR++ is one of my favorite tools.


SamRH:
This should resolve at least some of the instances of issues like #105 and #445.

PR #708, opened 2022-04-01
Allow user defined CMAKE_INSTALL_PREFIX

hb9fxq:
Please allow to use user defined CMAKE_INSTALL_PREFIX when building. Makes things easier when having to deal with several different build environments on a single machine.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,10 +1,12 @@
 cmake_minimum_required(VERSION 3.13)
 project(sdrpp)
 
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
Have you checked that the behavior remains unchanged when not providing an install prefix?
Are you sure this doesn't mean the values will never be overriden when the OS provides a different install prefix?

hb9fxq:
https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT.html

PR #711, opened 2022-04-02
Add SDDC support with SDDC Core Lib

howard0su:


AlexandreRouma:
Closing due to code quality and stability concerns.

As mentioned in other posts, support for the RX888 will not happen until a **real** user buildable library is available.

Issue #608, opened 2022-01-23
Feature request: mouse wheel zoom


Rua:
Can a feature be added to make the view zoom when the mouse wheel is rolled with the mouse over the FFT/waterfall, possibly with ctrl held (as in SDR#) or only when the mouse is over the frequency scale (as in Gqrx)? The view should zoom on the position of the mouse cursor, so whatever frequency is under the cursor should remain there and not move.

AlexandreRouma:
After reflection I won't be adding that feature. I scrolling (including with modifiers) will be reserved for other things.

PR #727, opened 2022-04-16
Move exponential zoom logic to WaterFall::setZoom

ericek111:
WaterFall::setZoom is declared in [waterfall.h](https://github.com/AlexandreRouma/SDRPlusPlus/blob/80d244003ad143a78ef1bd1cfd4a07c385baa77c/core/src/gui/widgets/waterfall.h#L155), but wasn't implemented.

I'm doing this in preparation of another PR.

### DIFF
--- a/core/src/gui/main_window.cpp
+++ b/core/src/gui/main_window.cpp
@@ -620,6 +620,7 @@ void MainWindow::draw() {
     ImGui::TextUnformatted("Zoom");
     ImGui::SetCursorPosX((ImGui::GetWindowSize().x / 2.0) - 10 * style::uiScale);
     ImVec2 wfSliderSize(20.0 * style::uiScale, 150.0 * style::uiScale);
+    bw = gui::waterfall.getZoom(); // check if the zoom hasn't been updated
     if (ImGui::VSliderFloat("##_7_", wfSliderSize, &bw, 1.0, 0.0, "")) {
         gui::waterfall.setZoom(bw);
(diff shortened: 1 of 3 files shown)


AlexandreRouma:
It's not the waterfall's job to handle the zoom controls, hence why it was never implemented.

I'd rather delete the declaration instead

ericek111:
I thought about that, but where else to put it? Is this more acceptable?

I want to implement scrolling with Ctrl+mwheel, as requested in #608, but I don't wanna duplicate the zoom algorithm.

AlexandreRouma:
You're forgetting that the zoom slider is supposed to update when the zoom is changed. That's why the zoom code is not part of the waterfall.
By the way, I'm not currently accepting feature pull requests since I don't have the time to validate the code.
Moreover quite a bit of the code is currently changing for the android UI compatibility.

ericek111:
I know, I realized. Forgive me for the abomination I have brought upon this world. Check out the code now.

This is the most convenient (for future development) way that came to mind. And it's not terrible. Or maybe it is. I don't know... I'm keeping it in my fork, it works fine. Some way to decouple the logic between WaterFall and MainWindow should be found.

ericek111:
`setZoom` still isn't implemented: https://github.com/AlexandreRouma/SDRPlusPlus/blob/50a77a7e608d59cdf1633f223b6900c8e51da351/core/src/gui/widgets/waterfall.h#L147

PR #728, opened 2022-04-16
cleanup: remove redundant search for hovered VFO

ericek111:
The `ImGui::IsMouseHoveringRect` check is done every frame for drawing the tooltip when hovering above a VFO. It's already done a few lines earlier at the beginning of `processInputs`, so there's no need to do it again.

I'm also debugging the `ImGui::GetIO().KeyCtrl` thing. It seems to be false when I start SDR++, then stays false as I hold Ctrl and becomes true upon releasing it (reversing the entire logic).

### DIFF
--- a/core/src/gui/widgets/waterfall.cpp
+++ b/core/src/gui/widgets/waterfall.cpp
@@ -442,38 +442,34 @@ namespace ImGui {
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


ericek111:
I guess spending a few CPU cycles extra won't hurt anyone...

PR #729, opened 2022-04-16
fix modifier keys on Linux by cherry-picking ImGui's 1ad8ad62

ericek111:
Anything with modifier keys is terribly broken in SDR++ (does the exact opposite and is buggy) on Linux/X11. This was fixed in https://github.com/ocornut/imgui/commit/1ad8ad623e7ec16f9ffb6b86ddf68bfd8deb43b9. I think rather than wait for a new release of ImGui, we should include that bugfix now.

### DIFF
--- a/core/backends/glfw/imgui/imgui_impl_glfw.cpp
+++ b/core/backends/glfw/imgui/imgui_impl_glfw.cpp
@@ -244,6 +244,19 @@ static ImGuiKey ImGui_ImplGlfw_KeyToImGuiKey(int key)
     }
 }
 
+static int ImGui_ImplGlfw_KeyToModifier(int key)
+{
+    if (key == GLFW_KEY_LEFT_CONTROL || key == GLFW_KEY_RIGHT_CONTROL)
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
Can't replicate on my X11 arch system (intel graphics).
I'd rather wait for the full ImGui 1.88 update than update partially.

ericek111:
**Steps to replicate:**
1. Start SDR++.
2. Hover over a VFO.
3. Press Control.
4. Release Control.

**Expected behaviour:**
Signal info should disappear, leaving only the VFO's name.

**Observed behaviour:**
Signal info stays on the screen after releasing the modifier key. As noted in the original commit, this is a bug -- that's why the commit exists.

**My machine:**
Arch Linux, MATE, X11, picom, SDR++ compiled from master.

AlexandreRouma:
Ah yeah guess that's broken.
Oh well, will wait for the 1.88 release since it's coming out soon anyway. Partial updates are never a good idea.

PR #738, opened 2022-04-24
waterfall: right click to tune the strongest signal

ericek111:
Right-clicking the waterfall tunes to the strongest displayed signal within 10 % of the waterfall where the user clicked. It uses FFT data from the waterfall display. Because of the limited resolution, it's not absolutely perfect (it's limited by resolution of FFT), but it's more than good enough. On 8 MHz bandwidth, I can comfortably listen to air band and click around without having to zoom in, adjust the VFO, zoom out to see the whole band...

I'm not sure if it should also snap according to the interval specified by the user, like with normal tuning using LMB.

### DIFF
--- a/core/src/gui/widgets/waterfall.cpp
+++ b/core/src/gui/widgets/waterfall.cpp
@@ -442,6 +442,11 @@ namespace ImGui {
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


AlexandreRouma:
hello,
I appreciate the pull request but I do not think a feature like this should be implemented directly in software. Feel free to create a module implementing this if enough people are interested.

PR #746, opened 2022-04-30
Frequency manager quick jump

OttoPattemore:
This pull request adds a popup dialogue to let users quickly jump between bookmarks.
The popup is activated by pressing **Ctrl + Space**.
It displays all bookmarks in the current list, with a search filter at the top.
The search filter is not case sensitive.
When a bookmark is selected the bookmark's frequency is tuned.
The popup dialog remains in case the wrong selection was chosen.
The popup is then closed by pressing **Enter** or **Escape** or by clicking **ok** in the popup.
image

### DIFF
--- a/misc_modules/frequency_manager/src/main.cpp
+++ b/misc_modules/frequency_manager/src/main.cpp
@@ -401,8 +401,7 @@ private:
         FrequencyManagerModule* _this = (FrequencyManagerModule*)ctx;
         float menuWidth = ImGui::GetContentRegionAvail().x;
 
-        if(ImGui::IsKeyPressed(ImGuiKey_Space)&&ImGui::IsKeyDown(ImGuiKey_LeftCtrl))
-        {
+        if (ImGui::IsKeyPressed(ImGuiKey_Space) && ImGui::IsKeyPressed(ImGuiKey_ModCtrl)) {
             _this->quickJumpOpen = true;
         }
 
--- a/misc_modules/frequency_manager/src/main.cpp
+++ b/misc_modules/frequency_manager/src/main.cpp
@@ -401,7 +401,7 @@ private:
         FrequencyManagerModule* _this = (FrequencyManagerModule*)ctx;
         float menuWidth = ImGui::GetContentRegionAvail().x;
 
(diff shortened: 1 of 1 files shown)


shuyuan-liu:
I remember Ctrl + Space being the key combo to change input methods (or keyboard layouts?) on Win 7, so this is potentially problematic for some users. Would you consider changing it to something else?

OttoPattemore:
Thanks.
Yes, better to use a different shortcut.
Perhaps **Ctrl + Q**.
Ctrl + Q is used by MS word, so it shouldn't conflict with any OS shortcuts.

shuyuan-liu:
On Windows probably yes. On Linux and especially with GTK apps, Ctrl + Q is commonly "quit application". Not unusable, but it will probably take some getting used to.

OttoPattemore:
Yes, I agree. Thanks for pointing that out.
**Ctrl + F** is used universally to search.
I couldn't find any consistently used shortcuts for searching bookmarks that also aren't used for other purposes.

misc_modules/frequency_manager/src/main.cpp
```diff
@@ -190,6 +190,58 @@ class FrequencyManagerModule : public ModuleManager::Instance {
...
+                    if (ImGui::Selectable((name + "##freq_manager_quick_jump_bkm_name" + this->name).c_str(), &bm.selected, ImGuiSelectableFlags_SpanAllColumns | ImGuiSelectableFlags_SelectOnClick)) {
+                        applyBookmark(bm, gui::waterfall.selectedVFO);
+                    }
+                    ImGui::TableSetColumnIndex(1);
+                    ImGui::Text("%s %s", utils::formatFreq(bm.frequency).c_str(), demodModeList[bm.mode]);
+                    ImVec2 max = ImGui::GetCursorPos();
+                }
+                ImGui::EndTable();
+            }
+            if (ImGui::Button("Ok",ImVec2(ImGui::GetContentRegionAvail().x,0))) {
```
victornpb, line 238:
```suggestion
            if (ImGui::Button("Close",ImVec2(ImGui::GetContentRegionAvail().x,0))) {
```
OK is usually used to confirm, since this button replaces the "X" I think it is more appropriate to call it "Close" instead.

Issue #701, opened 2022-03-24
Chinese characters are not supported and displayed“?”


long996:
Chinese characters are not supported and displayed“?”

shuyuan-liu:
Support for CJK (and other Unicode characters) is not planned. See #275.

AlexandreRouma:
after some imgui updates it might work if the font contained chinese characters, but unfortunately, it doesn't.

shuyuan-liu:
@AlexandreRouma would an option to select custom fonts (runtime or buildtime) be possible?

AlexandreRouma:
Not possible, ImGui relies on font size for scaling and the rest of the code doesn't support that

shuyuan-liu:
> Not possible

Tried replacing Roboto-Medium.ttf in the resource directory with Chinese fonts (Microsoft Sim Sun, WQY Micro Hei). The font does get used and nothing looks too broken; see screenshots attached. Would you mind testing / elaborating on why you think it's not possible? 

Sadly, with the new fonts, CJK characters are still `?`. The same `frequency_manager_config.json` as in #275 was used.

![Screenshot from 2022-03-27 19-48-56](https://user-images.githubusercontent.com/41851763/160296243-581d0190-39a4-4539-98e3-9851f1b08660.png)

![Screenshot from 2022-03-27 19-49-35](https://user-images.githubusercontent.com/41851763/160296248-ab05538f-a64f-4d39-8679-071b02409f6f.png)

AlexandreRouma:
changing the font isn't the only thing you have to do to make it work. You also have to tell ImGui to load the chinese characters.
Switch to the `backend_abstraction` branchy and add `builder.AddRanges(fonts->GetGlyphRangesChineseFull());` after line 30

bubnikv:
This is what we have done in PrusaSlicer to support wide range of languages:

This function initializes the glyph ranges.
https://github.com/prusa3d/PrusaSlicer/blob/master/src/slic3r/GUI/ImGuiWrapper.cpp#L130

Also based on the language we need to load the proper font. We currently distribute PrusaSlicer with two fonts for ImGUI rendering: NotoSans-Regular.ttf and NotoSansCJK-Regular.ttc, the latter used for Chinese, Japanese and Korean. The CKJ font is quite big.

[...]

PR #770, opened 2022-05-27
load all glyphs from the font to extend the displayed character range

### LINKED ISSUE #528: bookmark name does not support Unicode【1.0.4】
version：1.0.4

Bookmark names currently do not support Unicode characters, and characters such as CJK characters entered are displayed as ???

pinkavaj:
Allow display all the characters supported by the font.

Might close #528 and #701


AlexandreRouma:
this is not a valid way of doing things. ImGui doesn't check if the ranges actually exists and this can cause bugs when the missing characters show up.
The font ranges selected are all the ones provided by the font file used in the software

PR #814, opened 2022-07-23
Add UI control for vsync

luizribeiro:
On macOS, I have been experiencing some stuttering in the UI. This often happened when dragging the window or every few seconds while using SDR++, which would also cause the sound to stutter.

After debugging for a bit, it turned out it was due to vsync being enabled.

This adds an option in the UI for controlling vsync, as well as a setting on `config.json`. The setting defaults to enabled, so we keep the same default behavior as before (vsync enabled).

For now, there is no support for Android, as I don't have a device to test with - so this option is hidden on those builds.

### DIFF
--- a/core/backends/android/backend.cpp
+++ b/core/backends/android/backend.cpp
@@ -158,7 +158,7 @@ namespace backend {
         ImGui::NewFrame();
     }
 
-    void render(bool vsync) {
+    void render() {
         // Rendering
         ImGui::Render();
         auto dSize = ImGui::GetIO().DisplaySize;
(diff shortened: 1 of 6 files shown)


AlexandreRouma:
FIrstly, disabling VSYNC is just a really bad idea. Not only will the CPU and GPU usage spike up to barely usable levels, but it'll introduce screen tearing. Secondly, UI stutters no effect on the audio. The DSP is run in completely separate threads that only ever interact with the UI when settings are changed. These issues you are experiencing are more than likely random GLFW bugs.

luizribeiro:
I understand what vsync is :-)

> Not only will the CPU and GPU usage spike up to barely usable levels, but it'll introduce screen tearing

FWIW I didn't experience either. Would you be open to the idea of having this as a setting on `config.json` but not in the UI?

> These issues you are experiencing are more than likely random GLFW bugs.

You are probably right that this is glfw/glfw#857 but I would love to at least have a workaround.

AlexandreRouma:
Even if it only was a config file only setting, you can be sure people who try to debug their issue will try to change random stuff and that'll become one more variable for me to deal with when people report bugs. It's a setting that can only make things worse when enabled.

PR #815, opened 2022-07-24
Target macOS 10.15 when building

luizribeiro:
SDR++ relies on `std::filesystem` which is only available when targeting macOS 10.15 and above (which is what the CI is using).

SDR++ wouldn't build on 12.1 unless I specified this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,7 @@ project(sdrpp)
 
 if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
     set(CMAKE_INSTALL_PREFIX "/usr/local")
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
If you look around 20 lines under the modification, you'll notice:
```cmake
option(OPT_OVERRIDE_STD_FILESYSTEM "Use a local version of std::filesystem on systems that don't have it yet" OFF)
```
This allows to build SDR++ on much older systems (both Linux and MacOS) and I've had reports of it working fine even on High Sierra. (as mentioned in other issues I don't/have access to a mac so these are only user reports).

Users are expected to manually `-DCMAKE_OSX_DEPLOYMENT_TARGET=<whatever>` if their system requires it, which only appears to be a thing with the last updates apple did.

PR #822, opened 2022-07-27
Update readme.md

bkerler:
Make sure people don't use the latest 6.x version of rtaudio from github but the release 5.2.0 instead, as it will break compilation. (at least issues with rtaudioerrortype and probed)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -110,7 +110,7 @@ The preferred IDE is [VS Code](https://code.visualstudio.com/) in order to have 
 * [cmake](https://cmake.org)
 * [vcpkg](https://vcpkg.io)
 * [PothosSDR](https://github.com/pothosware/PothosSDR) (This will install libraries for most SDRs)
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
I'm not sure how to go forward. The issue is caused by a breaking change that I think should be reported to them before they release 6.0.
By the way, you only updated the windows section so people looking for linux or macos won't see anything.

bkerler:
Yes, I think I'm closing the issue for now. The 6.0 code on their github repo is now fully broken as they do not implement the .probed bool on the deviceinfo anymore. I wasn't able to find any replacement. To be honest I'm not sure if relying on rtaudio is a good idea regarding the state of the current rtaudio code. I'm now aware that using the latest rtaudio code is a bad idea. But you should at least somewhere add a warning regarding the rtaudio github repo for those who want to build from scratch (for example for appimages). Cheers.

PR #832, opened 2022-08-04
Create CODEOWNERS

uranuspucksaxophone:
This automatically requests PR reviews from @AlexandreRouma


AlexandreRouma:
Trust me I already get notified when there's a PR, I don't need more email spam...

PR #833, opened 2022-08-04
Optimize WaterFall::doZoom and fix imprecise zooming

ericek111:
The waterfall was jumping around when zooming, it made VFOs randomly (float imprecision) offset from signals displayed on the waterfall. I also optimized the `doZoom` inline method. Now it's roughly twice as fast.

### DIFF
--- a/core/src/gui/widgets/waterfall.cpp
+++ b/core/src/gui/widgets/waterfall.cpp
@@ -78,6 +78,7 @@ namespace ImGui {
         lastWidgetSize.y = 0;
         latestFFT = new float[dataWidth];
         latestFFTHold = new float[dataWidth];
+        tempZoomFFT = new float[dataWidth];
         waterfallFb = new uint32_t[1];
 
         viewBandwidth = 1.0;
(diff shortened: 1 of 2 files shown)


AlexandreRouma:
goddamn I didn't remember that part of the code being so horrible. Must have been one of these "4am" pieces of code...

Could you explain in more detail the mod to doZoom? I'd recommend commenting the code in that function a bit more.

ericek111:
I've added some comments, hopefully it's clear now. I tried googling a way to make it even faster, to no avail. I think the entire FFT should be moved to a separate thread anyway. I like having the full update functionality enabled and even with my 1080p screen and $25 SDR dongle, SDR++ lags when zooming fully out. Or, maybe, at the very least, the zoom slider could be debounced if full update is enabled.

AlexandreRouma:
The FFT is already in it's own thread. doZoom is called from the aquireFFTBuffer (or whatever don't remember the exact name) function which is called by the handler code. Only the ImGui rendering code runs in the main thread.
I'll check the code in detail when I get the time, extremely busy with university work at the moment.
Considering I'm really close to releasing 1.1.0 I need to garanteed validity and stability.

ericek111:
In that case, could you change the variables `factor,` `sFactor`, `uFactor` and `id` to doubles? That cannot break anything and it will fix the issue with waterfall jittering left and right.

AlexandreRouma:
> In that case, could you change the variables factor, sFactor, uFactor and id to doubles? That cannot break anything and it will fix the issue with waterfall jittering left and right.
I am 99% sure the issue is not float imprecision. it's simply the FFT bins being forces to stay aligned.
I will look into changing it to doubles since most of the other parts of the waterfall already use doubles.
Switching any variable representing a frequency or bandwidth is a good thing.

> Right, but the FFT zooming function runs in the main thread and doing the doZoom
That is wrong. The only case the FFT zoom is ever run in the main thread is in case full waterfall update is enabled, when switching color pallet or when the window is resized. (aka `updateWaterfallFb`). These is no reason to run it in another thread as it has to be synchronous.

AlexandreRouma:
open bug report instead

Issue #811, opened 2022-07-18
RTL-SDR serial number not displayed


scottjpalmer:
When using multiple RTL-SDR devices it is not immediately obvious which physical device is being selected. Displaying the device serial numbers in the drop down in the Source module would assist with this. Thanks.

benkyd:
Proposed fix in #853

AlexandreRouma:
implemented

PR #853, opened 2022-08-24
Add RTL-SDR Serial number to the device display

benkyd:
Fix for #811


AlexandreRouma:
Closed as implemented in a slightly different way

Issue #532, opened 2021-12-04
No way to enable bias-t for rsp1a running remotely through Soapy


cdbrinker:
Connects to remote computer hosting the rsp1a and pipes in the signal just fine but no way to enable bias-t. I can get strong FM stations poorly, the rest of the bands are just white noise. Did I build incorrectly? When running on the rsp++ machine with the SDRPlay source, no problems. CubicSDR has no trouble with Soapy on these two machines and even enables the bias-t by default.

AlexandreRouma:
The soapy source doesn't support bias-t remote or not. SoapySDR doesn't have a standard way of enabling bias-t.

AlexandreRouma:
Also I'll add that his sounds more like a gain issue, I doubt any SDR would have trouble receiving broadcast FM with its gain set correctly

cdbrinker:
> The soapy source doesn't support bias-t remote or not. SoapySDR doesn't have a standard way of enabling bias-t.

Not sure I understand. I can enable bias-t (or FM notch, DAB notch, etc) on CubicSDR via a remote Soapy server. So is there specific support in CubicSDR for my RSP1a that is not present in SDR++?

Just to clarify, the reception in SDR++ via Soapy remote is exactly the same on all bands as with bias-t disabled in CubicSDR via Soapy remote.

AlexandreRouma:
I suspect cubic has a list of SDRs and their option name to enable bias-t, then it uses that to enable or disable it. The soapy source in SDR++ is completely device agnostic. Not sure what I can do without manually checking for every device type which would defeat the purpose of the soapy source.

I think I might be able to add a custom device arg field to the soapy source, but you'll have to type in the right soapy args for it to turn on the bias-t

nmaster2042:
I don't think Cubic has a list of SDRs, everything is dynamic.

In Soapy driver, there is function to get device options (if avaliable), it's called getSettingInfo()
For SoapySDRPlay3 driver, line 1085 https://github.com/pothosware/SoapySDRPlay3/blob/master/Settings.cpp
But other drivers may have too with other options, depending of the hardware specifications, driver implementation.

And there is a writeSetting() line 1343 and readSetting() line 1533.
The issue I think is to generate device settings dynamically to provide it on the GUI.

Hope this can help

nmaster2042:
[...]

Issue #442, opened 2021-10-14
Expose more device options


troffasky:
I don't know if this only applies to Soapy Remote, but I cannot toggle the broadcast and DAB notch filters on my SDRplay RSP1a from sdrpp. CubicSDR exposes these options when using Soapy Remote.

troffasky:
Device string option would be a good start, wouldn't take much room in UI and doesn't require understanding of every device type.

AlexandreRouma:
Switch to the SDRplay source instead

PR #856, opened 2022-08-29
SoapySDR settings

ericek111:
Expose device configuration for SoapySDR drivers that implement the Settings API. All state is managed in the driver (through `readSetting` and `writeSetting`). Tested with dropdown menus, checkboxes, string inputs, int/float sliders with/without limits and/or steps.

Solves #532 and #442.

EDIT: float inputs without range are currently implemented as sliders with default range of 0-100. Ctrl+click can be used to input arbitrary values. `SmGui::InputFloat` should be implemented.

### DIFF
--- a/source_modules/soapy_source/src/main.cpp
+++ b/source_modules/soapy_source/src/main.cpp
@@ -210,6 +210,8 @@ private:
 
         hasAgc = dev->hasGainMode(SOAPY_SDR_RX, channelId);
 
+        settings = dev->getSettingInfo();
+
         SoapySDR::Device::unmake(dev);
 
         config.acquire();
@@ -248,6 +250,11 @@ private:
             else {
                 selectSampleRate(sampleRates[0]);
             }
+            if (config.conf["devices"][name].contains("settings")) {
+                for (auto& [key, value] : config.conf["devices"][name]["settings"].items()) {
+                    configSettings[key] = value;
+                }
+            }
         }
         else {
             uiAntennaId = 0;
@@ -279,6 +286,14 @@ private:
             conf["bandwidth"] = uiBandwidthId;
         if (hasAgc) {
             conf["agc"] = agc;
+        }
+        if (running) {
+            for (SoapySDR::ArgInfo argInfo : settings) {
+                std::string val = dev->readSetting(argInfo.key);;
+                if (val != argInfo.value) { // only save non-default values
+                    conf["settings"][argInfo.key] = val;
+                }
+            }
         }
         config.acquire();
         config.conf["devices"][devArgs["label"]] = conf;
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
Haven't tried it but I'm almost certain this is gonna duplicate setting and just make everything more confusing for no reason.
Not to mention the potential stability issues that would require testing with every SDR soapy supports which is just a waste of time.

The SDRplay SDRs should never be used with theh soapy source as they have their own dedicated source module so I'm closing both issues mentioned.

Additionally, SoapySDR support will likely be depracated by the end of the year once an audio source and USRP devices are supported (should happen next month). It's been a constant source of instability and confusion since people keep trying to use it when native sources are available.

If you have an SDR that isn't natively supported, name it and I'll try to add native support for help from the manufacturer. If you're using it with SoapyRemote, consider using SDR++ server instead which is better in every way imaginable.

ericek111:
I have an RSP1 clone supported by libmirisdr using my own driver. Soapy has been nothing but flawless for me. On the other hand, the buggy unstable proprietary piece of shit SDRplay's "API" was a constant source of headache. Always had to restart it 5 times before it would work (same experience across all OSs), needs root to work, is limited to x86 and Raspberry Pi, it's closed source... If I bump my SDR dongle, close SDR++, restart the service, restart one more time, try SDR++, refresh devices, crashes SDR++, restart the service, restart SDR++... Libmirisdr through my SoapyMiri module made my life SO much easier (I can have multiple SDR programs running, even across operating systems in VM with USB passthrough, the last app started takes control of the dongle). Also, SDRplay is dropping support for RSP1.

With my SoapyMiri driver, it duplicates nothing, as there is no reason for any gain or bandwidth elements to be exposed using the Settings API. It gives a checkbox to disable the automatic offset tuning and enable/disable bias T.

I believe that Soapy is the answer to complete interoperability between vendor-specific libraries and I/Q-taking apps. Dropping support in favour of handwritten libraries specific for every device would be a mistake. I would say, if you wanna waste your time on writing code that's already been written, focus your attention on audio output modules, they're buggy as hell (which I think is the fault of the underlying libraries). But removing the Soapy source module is not the right step, in my opinion.

AlexandreRouma:
> I have an RSP1 clone

Well that's your first problem. That clone is known to not exactly be the most reliable SDR out there...
The SDRplay API is unstable on linux, we all know that, but that's something for SDRplay to fix.

> I believe that Soapy is the answer to complete interoperability between vendor-specific libraries and I/Q-taking apps.

SoapySDR has consistently been a source of headache for years. It's constantly causing conflicts (resulting in complete crashes or weirder issues) with the native sources (RTL-SDR, HackRF, RSPs, and more), it slows down the start up of the app significantly and adds a layer of abstraction which measurably decrease performance. 

I am not wasting my time by "writing code that already exists", because it **doesn't** already exist. Native source modules allow a tailored interface for each SDR with maximal performance without relying on anyone else's code.

If you can write your own Soapy module, you can probably write your own SDR++ source module as well. I'll note that there are existing third party MiriSDR source modules availables.

PR #888, opened 2022-10-14
fixed error D8016 for MSVC debug build

KuzVlad:
After generating with cmake .sln for Visual Studio and compiling it, the following error will be received:

image

In short, the optimization setting Maximize Speed (/O2) is incompatible to the Basic Runtime Checks setting of Code Generation.

This problem can be fixed through the project configuration in Visual Studio:
**Solution->Properties->Configuration Properties->C/C++->Code Generation->Basic Runtime Checks** and set it to **"Default"**

image

But I think it's more logical to separate the Debug and Release MSVC compiler arguments in the cmake project.


AlexandreRouma:
Should be fixed in the latest commit but you need to make sure to set `CMAKE_BUILD_TYPE` to `Debug`

PR #896, opened 2022-10-23
Updated rpi-install.sh and read.md to use libvolk2-dev

HuggeK:
Tested to run the shell script on a raspberry pi 3b+ and libvolk1-dev did not seem to work. Switched to libvolk2-dev and it did build!


AlexandreRouma:
it depends on which version of the pi image it's built for. Existing pi3 users likely run an older image which won't have volk2

HuggeK:
Ok

PR #915, opened 2022-11-15
Correct USA amateur allocations below 2GHz

jeffrizzo:
- add 2200m band
 - add 630m band
 - fix 80m band - in US, it's 3.5-4.0 MHz
 - 60m band is 5 discrete "channels"
 - 40m band in US is 7.0-7.3 MHz
 - 10m band in US is 28.0-29.7 MHz
 - 1.25m band is in two parts; 219-220MHz and 222-225 MHz

Source: ARRL, FCC

### DIFF
--- a/root/res/bandplans/usa.json
+++ b/root/res/bandplans/usa.json
@@ -6,6 +6,18 @@
     "author_url": "http://jkimsey.tech",
     "bands": [
         {
+            "name": "2200m Band",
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
band plans are not meant to have "discrete channels", they show the bands used for certain purposes. Otherwise CB would have dozens of individual band plans making it hard to read.

Issue #718, opened 2022-04-08
SDRPP Conflicts with WineHQ dependencies when installing using dpkg/apt


SpaceXCheeseWheel:
Attempting to install SDR++ using the deb files on the latest version of Linux Mint attempts to uninstall several packages such as WineHQ, libasound, and libjack:

```
libasound2-plugins:i386
libjack-jackd-2-0
libjack-jackd-2-0:i386
wine-devel
wine-devel-i386:i386
winehq-devel
winetricks
```
I am using `sdrpp_ubuntu_focal_amd64.deb` from the most recent release.

![image](https://user-images.githubusercontent.com/12452268/162351751-88b6ca80-3df9-44aa-ba38-9bf6b9ddd07a.png)

From what I understand, the source of the problem is `libjack0`. `libjack-jackd2-0` is a dependency of wine, and `libjack0` is one of SDR++s' dependencies. According to apt, `libjack0` conflicts with `libjack-jackd2-0`. Apt tries to uninstall libjack-jackd2-0, and it just works up the chain. I could be completely wrong about this, though. 

I am unsure exactly how to troubleshoot this any further. For anyone trying to replicate this, I am using wine packaged by WineHQ following [this guide](https://wiki.winehq.org/Ubuntu). In other words, I am not using wine from the official ubuntu repositories.

AlexandreRouma:
either you got the wrong deb or your distro's not up to date or your graphical package manager is broken.
De b files cannot ask to remove packages. It can only ask to install some

SpaceXCheeseWheel:
Reproduced using apt, but seems to be exclusive to my current configuration, was not able to reproduce on another Linux mint system. Closing for now.

RealJTG:
> either you got the wrong deb or your distro's not up to date or your graphical package manager is broken. Deb files cannot ask to remove packages. It can only ask to install some

Actually apt [can](https://askubuntu.com/questions/1308117/ubuntu-20-04-installing-wine-has-downgraded-packages-cant-see-new-versions) [ask](https://unix.stackexchange.com/questions/154977/why-is-apt-get-trying-to-remove-certain-packages-when-i-install-wine) [for](https://forum.winehq.org/viewtopic.php?f=8&t=35440) [removal](https://community.mycroft.ai/t/there-should-be-a-warning-that-mycroft-will-uninstall-wine-and-other-software/6995), it isn't the smartest package manager, sadly.

[...]

PR #919, opened 2022-11-19
Improvements for Debian builds

okaestne:
This PR improves the generation of the Debian packages that are build by the GitHub workflow:

* use `apt-get` instead of `apt`
  * `apt-get` should be preferred for scripts
* use `nproc` threads for make
  * to speed up (docker) builds, if used locally (GitHub Linux runners only use two threads, so no improvement there)
* use non-`dev` package dependencies
  * **users** don't need to install library headers and static libraries.
  * this will prevent the issue described in #718
  * this unfortunately requires more specific (and distribution-version-dependent) package names including  (e.g. `libvolk2.2` on focal, `libvolk2.5` on jammy), so a bit harder to maintain.
* adjusted workflows:
  * add distribution codename (e.g. bullseye) and architecture in .deb file name automatically
  * apply `shellcheck` recommendations to scripts

~I will need to test this on all distros, so marked as WIP~. I would appreciate help and feedback. Also, I can update the build instructions for Debian/Ubuntu users in another PR, if you wish.

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -143,13 +143,13 @@ jobs:
 
         - name: Recover Deb Archive
           working-directory: ${{runner.workspace}}
-          run: docker cp build:/root/SDRPlusPlus/sdrpp_debian_amd64.deb ./
+          run: docker cp build:/root/SDRPlusPlus/sdrpp_buster_amd64.deb ./
 
         - name: Save Deb Archive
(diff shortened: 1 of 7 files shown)


okaestne:
Rebased and successfully tested on fresh buster, bullseye, focal and jammy VMs with my RTL-SDR.

AlexandreRouma:
Please open a bug report instead

PR #934, opened 2022-12-10
Added MMDS converter offset preset

RAD750:
This pull request adds a preset for a 1998 MHz local oscillator MMDS converter, popular for use in S-band satellite reception.


AlexandreRouma:
Could you put it in the right place in terms of frequency order?

RAD750:
Sure! Also added the missing comment and renamed it slightly because it didn't fix in the box :)

PR #945, opened 2022-12-25
Fixed "the stream is not open" error message

CircuitChaos:
The exact error message is:

terminate called after throwing an instance of 'RtAudioError'
  what():  RtApiPulse::stopStream(): the stream is not open!
Aborted

### DIFF
--- a/sink_modules/audio_sink/src/main.cpp
+++ b/sink_modules/audio_sink/src/main.cpp
@@ -186,8 +186,10 @@ private:
         stereoPacker.stop();
         monoPacker.out.stopReader();
         stereoPacker.out.stopReader();
-        audio.stopStream();
-        audio.closeStream();
+        if (audio.isStreamRunning())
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
Hi,

Thanks for the bugfix!
I think a better way to implement this fix would be to actually go to the root issue:
I forgot to save whether the stream is successfully started or not. This can be fixed by adding a variable that remembers that state and checking it before start/stop

I'll edit the PR and merge it in a second.

CircuitChaos:
Hi,

Thanks for the quick reply. You're right, it would be the better this way.

AlexandreRouma:
Could you confirm the issue has been fixed with commit [c0c5b11]?(https://github.com/AlexandreRouma/SDRPlusPlus/commit/c0c5b1186cb91e04868d863d404c06abf7695167)

PR #963, opened 2023-01-12
Adding Fedora Linux Install info

bogomil:
There is a package available. Tested and working on Fedora 37.


AlexandreRouma:
The package is missing half of the modules so no, I don't recommend it.
Build from source instead.

PR #981, opened 2023-02-02
Add volume control

Georg1775:
Volume slider in Recorder was only affecting the bar graph but has no influence to the recorded wav file. In case of e.g. recording an USB signal the volume was always too high, causing clipping and distortion in the wav file.


AlexandreRouma:
It's a bug but that's not the correct way to fix it, it's just reading from the wrong stream.

Open a bug report issue instead, I'll update when it's fixed.

AlexandreRouma:
FYI it was fixed yesterday

PR #997, opened 2023-02-24
RDS Radiotext conversion on the fly from EBU charset to UTF-8

es5nhc:
This patch makes the RDS Decoder output RadioText content in UTF-8 instead of the EBU charset used for radiotext content in the RDS specification.

The conversion to UTF-8 should eliminate decoding problems experienced in countries on stations that use non-ASCII characters(ä, ü, š etc.) in their radiotext. Any such characters are encoded on the raw broadcast using an 8-bit code-table defined by the EBU, which won't show up properly on computers, which likely expect UTF-8 instead. I've used this patch successfully on Kubuntu 22.04. 

**How the conversion is made**
The conversion is only done when the Radiotext content is requested using getRadiotext() in RDSDecoder class. The radiotext string buffer is iterated through letter-by-letter with a conversion map. The converted letters will be appended to a separate string which will become the string returned by getRadioText(). This approach leaves the radiotext buffer untouched so that it can be safely updated with new data coming in over the air.

Example of patch in action:
Screenshot_20230224_183954


AlexandreRouma:
Hi,

This shouldn't be submitted here, the RDS library is at https://github.com/AlexandreRouma/rds.

In terms of the code, I'm pretty sure there's a more elegant and efficient way of doing the conversion than one huge const map.

By the way, I no longer accept feature pull requests, so I suggest opening an issue in the RDS library repo instead.

es5nhc:
Thank you for the info. I will open an issue at the referenced repository.

undefinedactive:
Still no updates?
This is the issue:
https://github.com/AlexandreRouma/rds/issues/1

PR #1017, opened 2023-03-21
flog.cpp: Honor current background colour

gvanem:
The removal of spdlog left the console-log with only black background colours on Windows.
I normally have a blue background. This patch OR's the new foreground over the background.


gvanem:
Re. 2nd part of the PR. Using hi-intensity colours is IMHO better. From this:
sdr++-console-1

To this:
sdr++-console-2

AlexandreRouma:
Open an issue on the correct repository: https://github.com/AlexandreRouma/flog
PS; Just setting it to bright is wrong too, it has to be set to whatever it detects during initialisation.
The windows terminal is downright dumb for this kind of stuff...

PR #1022, opened 2023-03-23
Fix LimeSDR Mini v2 radio start

sergeyvfx:
This fixes the issue when starting LimeSDR Mini v2 radio does not really start the actual reception. The workaround was to configure the bandwidth after the radio was started.

The actual fix is to enable TX channel. This is how the demo code from LimeSuite is communicating with the v2 of the radio, and this is also was confirmed on the myriad discourse:

  https://discourse.myriadrf.org/t/limesdr-mini-v2-reading-samples-from-rx-fifo/7900/2

The TX channel is only enabled for the LimeSDR Mini v2 and not for other LimeSDR radios to avoid unwanted possible side effects. The check is performed based on the name, matching the LimeSDRTest application from the LimeSuite. The name of the board is inlined to avoid dependency of the new LMSBoard.h.

Upon the radio stop both RXC and TX channels are stopped. This is for the code simplicity. Alternative would be to rely in the LMS_Close() to disabled channels which matches the basicRX demo code from the LimeSuite. It is not really clear what is the best strategy here.

### DIFF
--- a/source_modules/limesdr_source/src/main.cpp
+++ b/source_modules/limesdr_source/src/main.cpp
@@ -7,6 +7,7 @@
 #include <config.h>
 #include <gui/smgui.h>
 #include <lime/LimeSuite.h>
+#include <string.h>
 
 
 #define CONCAT(a, b) ((std::string(a) + b).c_str())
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
This is a bad way to do it, you're gonna be wasting CPU and USB bandwidth for users who are using other devices than the Lime mini 2.0.

I'll fix it myself properly when I get time to work on the mini v2 stuff, open an issue.

sergeyvfx:
Not sure where the waste will be coming from. From my understanding this is only enabling some internal registers. The actual data stream is still only opened for reception only. The transmit LED status on the device was showing OFF, and could not see any extra transfers in the libusb logs. Kind of confusing, so I've asked for clarification in the discourse.

Anyway, filed an issue here #1025.  Hopefully it explains problem well enough.

AlexandreRouma:
Alrighjt I'll keep an eye on the forum as well.
It seems the v2 support in limesuite still needs some work ngl...

PR #1027, opened 2023-03-27
Fix installation directories

hamarituc:
The library path is hardcoded to `/usr/lib` (or `/usr/local/lib` depending on the chosen prefix). But depending on the architecture, libraries must be installed in `/usr/lib64`. CMake provides variables which handle this case discriminations. This PR replaces the hardcoded values by these variables.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,10 @@
 cmake_minimum_required(VERSION 3.13)
 project(sdrpp)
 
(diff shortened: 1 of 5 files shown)


AlexandreRouma:
I've never seen a single linux distro where this matters and personally don't care much for the goofy name standards that some distros have done (why would the directory with the native binaries have the non-standard suffix...).

PR #1041, opened 2023-04-09
Fix jack audio repeating bug

atlas-devel:
If you run SDRPlusPlus with jack using the regular audio_sink module and stop the playback, the audio keeps repeating as the buffer is never cleared. This does not occur under portaudio.
The only possible issue with this fix is that it keeps clearing the buffer over and over again, but the portaudio sink module does this too.

### DIFF
--- a/sink_modules/audio_sink/src/main.cpp
+++ b/sink_modules/audio_sink/src/main.cpp
@@ -9,7 +9,7 @@
 #include <RtAudio.h>
 #include <config.h>
 #include <core.h>
-
+#include <string.h>
 #define CONCAT(a, b) ((std::string(a) + b).c_str())
 
 SDRPP_MOD_INFO{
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
I don't like how the portaudio sink does it either, with the new sink system coming outputting audio while the SDR is stopped should be explicitely supported (eg. for audio alerts or accessibility features like setting readout).

I'll keep this open for now but I think submitting an issue to RtAudio to have more consitent behavior over different backend. JACK seems to be the only backend that has this issue.

atlas-devel:
The best fix would be for RtAudio to close the jack client when you stop the SDR, and restart it, when you start it again, since that's where the main issue lies

AlexandreRouma:
Re read what I have said:
> with the new sink system coming outputting audio while the SDR is stopped should be explicitely supported (eg. for audio alerts or accessibility features like setting readout).

If you close the jack client, you can't output audio. This cannot be coupled with the play state of the SDR.

atlas-devel:
Ah, sorry, somehow I didn't realize that!
Would need to find a way to clear the buffer once when we stop the SDR, but not keep clearing it over and over again then

AlexandreRouma:
Alright so in any case the rewrite of the sink system will make this PR unusable.
Please open a bug report issue instead.

PR #1067, opened 2023-05-14
Add missing includes

ericek111:
My branch fails to compile on latest Arch with two errors. Your master compiles fine, but better do this to prevent some possible future confusion.
```
SDRPlusPlus/core/src/utils/net.cpp:71:24: error: ‘runtime_error’ is not a member of ‘std’
   71 |             throw std::runtime_error("Unknown host");

SDRPlusPlus/core/src/gui/smgui.h:89:49: error: ‘uint8_t’ has not been declared
   89 |         static int loadItem(DrawListElem& elem, uint8_t* data, int len);
```

EDIT: Wait, maybe I'm retarded.

### DIFF
--- a/core/src/gui/smgui.h
+++ b/core/src/gui/smgui.h
@@ -1,6 +1,7 @@
 #pragma once
 #include <stdint.h>
 #include <imgui.h>
+#include <stdint.h>
 #include <string>
 #include <vector>
 #include <map>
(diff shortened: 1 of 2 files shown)


AlexandreRouma:
remove the net one it's already been fixed in the upstream repo, just not updated

ericek111:
After pulling from master all works fine now. Weird (the stdexcept in net.h is still missing). Oh well, honestly sorry for wasting your time, should've pulled first.

AlexandreRouma:
checked the master branch and both those includes are already there so might wanna git reset

PR #1115, opened 2023-07-02
feat: allow several fractional scaling options

DuMOHsmol:
Hi, thanks for the great software!

I've got a laptop with 14" FullHD display. 100% scale is too small for it, and 200% is too huge. So, I've decided to add several fractional scaling options. They seem to work well enough, got no glitches while playing around with my RTL-SDR.

### DIFF
--- a/core/src/gui/menus/display.cpp
+++ b/core/src/gui/menus/display.cpp
@@ -114,10 +114,13 @@ namespace displaymenu {
         updateFFTSpeeds();
 
         // Define and load UI scales
-        uiScales.define(1.0f, "100%", 1.0f);
-        uiScales.define(2.0f, "200%", 2.0f);
-        uiScales.define(3.0f, "300%", 3.0f);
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
There is a good reason why those resolutions are not available. They cause some widgets to be misaligned. Until these things are fixed I cannot enable those scaling options.

DuMOHsmol:
Got it, thanks. Shall I open the issue for it then? I could not find anything related to fractional scaling in the bug tracker.

And which widgets are broken exactly? Maybe I could try checking them as well.

AlexandreRouma:
Create an issue yes. Don't bother with trying to fix the widgets. Hacking all of them to work properly is not a solution, it needs to be a more general change to how scaling is done in the first place.

DuMOHsmol:
Ok, created it: https://github.com/AlexandreRouma/SDRPlusPlus/issues/1116

PR #1152, opened 2023-07-28
gh workflow: fix build on Debian sid (current 13) libvolk-dev now is 3.0 ver

Mr-Precise:
If I understand correctly, libvolk-dev in Debian Sid (currently) is version 3.0 https://packages.debian.org/sid/libvolk-dev
and libvolk2-dev is no longer available
Replace libvolk2-dev to libvolk-dev


AlexandreRouma:
Fixed

PR #1163, opened 2023-08-06
fixed buffer overflow in file source

theverygaming:
This PR fixes
- Buffer overflow when the supplied WAV has a samplerate higher than STREAM_BUFFER_SIZE * 200 (which is very much possible - STREAM_BUFFER_SIZE * 200 is 200M while WAV allows for a samplerate of ~4294M)
- Crash when a WAV with samplerate = 0 is supplied. This causes buffers of size zero to be allocated which obviously results in a crash

These two bugs seem to be the only cause of SDR++ crashing when loading invalid WAV files

### DIFF
--- a/source_modules/file_source/src/main.cpp
+++ b/source_modules/file_source/src/main.cpp
@@ -1,3 +1,4 @@
+#define NOMINMAX
 #include <imgui.h>
 #include <utils/flog.h>
 #include <module.h>
--- a/source_modules/file_source/src/main.cpp
+++ b/source_modules/file_source/src/main.cpp
@@ -12,14 +12,6 @@
 #include <algorithm>
 #include <stdexcept>
 
-// TODO: figure out where exactly these macros are from (only happens on windows so probably from Windows.h somewhere)
-#ifdef min
-#undef min
-#endif
-#ifdef max
-#undef max
-#endif
-
 #define CONCAT(a, b) ((std::string(a) + b).c_str())
 
 SDRPP_MOD_INFO{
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
change it to reject the file if the SR is null. There is no point to even attempt to play it.

theverygaming:
done

theverygaming:
@AlexandreRouma

AlexandreRouma:
better but what's all that macro junk at the top?

theverygaming:
Something included creates min and max macros which interfere with std::min and std::max...
Probably Windows.h as mentioned in the comment. Defining the macro ``NOMINMAX`` before including Windows.h would be sufficient.

I can make it a bit more clean i guess. But it would be good to find the actual source of this problem and fix it in the relevant file since it will definitely cause headaches at some later point

AlexandreRouma:
a lot of other places use std::min/std::max just fine so this doesn't make much sense.

theverygaming:
Why doesn't it make much sense? Windows.h is not included a lot in SDR++ but when it is most of the time it's without the NOMINMAX macro defined..

There are various places where this macro could be coming from.
https://github.com/AlexandreRouma/SDRPlusPlus/blob/master/core/src/module.h#L18
https://github.com/AlexandreRouma/SDRPlusPlus/blob/master/core/src/gui/file_dialogs.h#L19
https://github.com/AlexandreRouma/SDRPlusPlus/blob/master/core/std_replacement/filesystem#L157

in this case i'm pretty sure it's from module.h

I do not have a windows build enviroment for SDR++ set up so it is hard for me to test this - i came across this issue when the windows build broke in CI..

AlexandreRouma:
interesting so my local env is immune to this.

PR #1164, opened 2023-08-06
Add RDS charset table

attah:
Add the charset lookup table (aka repertoire) for RDS.
More than the defined control codes in the lower two sections seems to appear here; so putting them all as space instead of trying to decide which should look broken vs not.


AlexandreRouma:
Alright so, this should normally be PRd on the RDS library repo () but anyway:
1) Do not use wstring, ImGUI expects UTF8 strings instead. Converting to wstring and back is just a waste.
2) For code clarity, use an array to define all the translation instead of a string.

attah:
> Alright so, this should normally be PRd on the RDS library repo 

/me mumbles about submodules. Will move once we have agreed on a solution.
Since it spans over two repos, rds being broken out and wfm being here(?), it's kind of impossible to work in the same structure that i would deliver.

But anyway...

> 1. Do not use wstring, ImGUI expects UTF8 strings instead. Converting to wstring and back is just a waste.

I'm really only going *forward* from wstring to utf8.
The lookup table is wchar-array, made with a wchar string constant.

While a string can contain utf8 data, it isn't really distinct from a plain string.
The problem i am working around is that i cannot index into programServiceName/radioText and place utf8 characters at their correct positions since utf8 characters have variable size.
I need some indexable representation (i.e. with fixed char-size) to be able to assemble the message, and wstring fits the bill.
I agree it looks a bit expensive; but i can't see what an alternative would look like.

> 2. For code clarity, use an array to define all the translation instead of a string.
Sure, why not. I guess i'm a bit C/C++ damaged.

AlexandreRouma:
> The problem i am working around is that i cannot index into programServiceName/radioText and place utf8 characters at their correct positions since utf8 characters have variable size.

Simply keep track of where each part of the string starts.

btw rename that array `EBU_TO_UNICODE`, that's more descriptive.

Also you deleted the content of a the `create_root.sh` file

AlexandreRouma:
Screw that I'll just implement things myself, thanks for wanting to help but going back on forth on issues just takes more time than doing it myself. I should have stuck to my policy of not accepting feature PRs

attah:
> Also you deleted the content of a the create_root.sh file

I did no such thing.
It is a permission change; to make it actually executable like it is supposed to be.
Tbh i don't recall doing that, but sounds reasonable. Though for a separate commit.

AlexandreRouma:
ok idk but github shows it as empty for some reason
image

Issue #893, opened 2022-10-20
[Feature Request] recording file name templates


thomasdouwes:
A text box or config file entry in the record module allowing you to enter a file name template, the current template could be the default as
`%(type)_%(frequency)_%(hour)-%(minute)-%(second)_%(day)-%(month)-%(year).wav`
or you could use a largest to smallest date with 
`%(type)_%(frequency)_%(year)-%(month)-%(day)_%(hour)-%(minute)-%(second).wav`
or something different like
`%(year)-%(month)-%(day)_%(hour)-%(minute)-%(second)_%(type)_%(frequency).wav`
or even subdirectories with
`%(year)-%(month)-%(day)/%(type)_%(frequency)_%(year)-%(month)-%(day)_%(hour)-%(minute)-%(second).wav`
this would be useful for organising recordings without manually sorting them
I have had a go at this and got it mostly working except for %(type) and subdirectory's, my code is awful but i could submit a pull request after tidying up my code

AlexandreRouma:
already planed but not gonna be as verbose, it has to fit in the side menu.

romanremus:
hello everyone......can you please compile sdr++ also for windows 7 /64 bit.....or please provide support in the source code for compiling in cygwin...........I have I tried to open it with windows 7 but I don't have SoapySDR.dll.....and in compilation with cygwin...I don't have certain dependencies and I can't install them in cygwin

thomasdouwes:
Closing because of name template in new recorder

rs1729:
Is it possible to add radio demodulation type ("NFM", "WFM", "AM", ..., "RAW") to "Audio" recording filename?

AlexandreRouma:
For the record this was implemented

PR #1166, opened 2023-08-18
Add demodulation mode to recorder filename template

rs1729:
https://github.com/AlexandreRouma/SDRPlusPlus/issues/893#issuecomment-1484538632

Recorder filename

If $r is used in name template, it adds the demodulation type of "Radio" to the filename.
For "Baseband" it would be "IQ",
for "Audio" one of
"NFM", "WFM", "AM", "DSB", "USB", "CW", "LSB", "RAW".

(If radio "RAW" is selected, recording "Stereo" is recommended.)

### DIFF
--- a/misc_modules/recorder/CMakeLists.txt
+++ b/misc_modules/recorder/CMakeLists.txt
@@ -5,4 +5,4 @@ file(GLOB SRC "src/*.cpp")
 
 include(${SDRPP_MODULE_CMAKE})
 
-target_include_directories(recorder PRIVATE "src/")
+target_include_directories(recorder PRIVATE "src/" "../../decoder_modules/radio/src")
--- a/misc_modules/recorder/src/main.cpp
+++ b/misc_modules/recorder/src/main.cpp
@@ -21,6 +21,7 @@
 #include <core.h>
 #include <utils/optionlist.h>
 #include <utils/wav.h>
+#include <radio_interface.h>
 
 #define CONCAT(a, b) ((std::string(a) + b).c_str())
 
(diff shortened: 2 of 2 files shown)


AlexandreRouma:
Many things wrong here:
1) Why get the selected VFO when the name is based on the **recorded** VFO and its name is already given (see line 448 for example).
2) Why a huge if/else statement when `std::map<int, const char*>` exists?
I'll implement this myself

AlexandreRouma:
Seems that the huge switch statement is because it was copied from the rigctl_server code, my bad. I'll fix that too, it's unacceptable/

rs1729:
Sorry, I don't know the code very well. Yes, I believe I took rigctl_server as an example, and the if/else style. There is another module, discord_integration
https://github.com/AlexandreRouma/SDRPlusPlus/blob/9c1361a8a9b5f2d41d84507adb7c39b41d69c7cf/misc_modules/discord_integration/src/main.cpp#L84-L109
that is very similar.

My idea was: If recording baseband, append "IQ", else if recording the selected VFO audio, append the selected mode.
"RAW" audio would be like "IQ", but only the 48k IF. Also I was wondering if "stereo" should be default for RAW?

Thanks.

rs1729:
Just tested, now if I record baseband, I get e.g.:
`baseband_20230906_114527_403500066Hz_Unknown.wav`

Would rather set
https://github.com/AlexandreRouma/SDRPlusPlus/blob/9c1361a8a9b5f2d41d84507adb7c39b41d69c7cf/misc_modules/recorder/src/main.cpp#L470
to `modeStr = "IQ"`,
such that
`baseband_20230906_114527_403500066Hz_IQ.wav`.
"baseband"..."IQ" is a bit redundant, but the default if radio/audio is not selected.

Another question:
What if the returned `mode` in
```
            core::modComManager.callInterface(name, RADIO_IFACE_CMD_GET_MODE, NULL, &mode);
            modeStr = radioModeToString[mode];
```
 is not found  in `radioModeToString[mode]`?

PR #1176, opened 2023-09-06
Fix CI for macOS and build Osmocom RTL-SDR drivers

pedroCX486:
## What this does?

- Fixes GitHub Actions for macOS builds by removing useless packages from Homebrew that can/may/will break the build environment

- Adds a step to fetch and build the latest Osmocom RTL-SDR Drivers, thus adding RTL-SDR Blog v4 support, instead of relying on homebrew packages that take ages to be updated

### DIFF
--- a/.github/workflows/build_all.yml
+++ b/.github/workflows/build_all.yml
@@ -93,7 +93,7 @@ jobs:
         - name: Create Build Environment
           run: cmake -E make_directory ${{runner.workspace}}/build
 
-        - name: Remove packages that may break brew
+        - name: Remove ununused brew packages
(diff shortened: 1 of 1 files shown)


pedroCX486:
> Thanks but no need, already been working on it today.

[365ab23](https://github.com/AlexandreRouma/SDRPlusPlus/commit/365ab2325e4107c042b874a127b4d4bb06e5ad61) is the wrong fix for it. You're avoiding homebrew updating the packages by doing that.

Did you at the very least read my PR?

AlexandreRouma:
*Your* "fix" is wrong. Why add a useless step when simply *removing* a work around for a *previous github CI bug* fixes the issue. The very CI run you linked completed successfully.

PR #1191, opened 2023-09-17
draws booksmarks on severl rows

daviderud:
Small mod to show waterfall bookmars on multiple rows (when shown on TOP).

image

### DIFF
--- a/misc_modules/frequency_manager/src/main.cpp
+++ b/misc_modules/frequency_manager/src/main.cpp
@@ -33,7 +33,12 @@ struct WaterfallBookmark {
     std::string listName;
     std::string bookmarkName;
     FrequencyBookmark bookmark;
+    float xmin;
+    float xmax;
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
As noted in the PR template, feature pull requests are not accepted.
Plus I've said several times that I don't want staggered bookmarks since it completely obstructs the view of the FFT...

daviderud:
OK I misunderstood the template, its says "Pull requests adding features or any bug fix that requires significant code changes will be automatically rejected". Was not sure about the "significant code changes" part (understood as "adding features without significant code changes could be accepted"). Propose changing into: "Pull requests adding features will be automatically rejected. Pull requests for any bug fix that requires significant code changes will be automatically rejected."

rgds

AlexandreRouma:
I agree this is grammatically ambiguous, I will update it.

Issue #958, opened 2023-01-07
Higher bandwidth for demodulators


kord1e:
**Feature description**

Currently SDR++ allows quite limited bandwidth for AM/SSB/DSB demodulators. It's often enough but sometimes there are signals that would benefit from higher bandwidth. For example, Gqrx allows higher bandwidth for these modes compared to SDR++.

Are there disadvantages if higher bandwidth is allowed?

AlexandreRouma:
> Are there disadvantages if higher bandwidth is allowed?

Wasting CPU usage by running the demodulators at a higher samplerate for no reason or dramatically increasing complexity and and decreasing responsiveness by dynamically changing the samplerate at which the demodulators run?

There is no reason to have the modes go wider than actually needed because they are configured for normal usage. If you're trying to receive a complete out of spec signal it likely makes much more sense to have a dedicated demodulator.

kord1e:
@AlexandreRouma 
Ok, I see. But SDR++ already performs better than other SDR software so is it much resources wasted? :P

PR #1203, opened 2023-10-03
Increase bandwidth to 60k

spithash:
Some amateur AM middle wave stations are much wider than 15k. It's nice to have the range to widen it.

### DIFF
--- a/decoder_modules/radio/src/demodulators/am.h
+++ b/decoder_modules/radio/src/demodulators/am.h
@@ -73,7 +73,7 @@ namespace demod {
         // ============= INFO =============
 
         const char* getName() { return "AM"; }
-        double getIFSampleRate() { return 15000.0; }
+        double getIFSampleRate() { return 60000.0; }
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
As already discussed *several* times...
https://github.com/AlexandreRouma/SDRPlusPlus/issues/958

spithash:
So I guess those who need it must create and/or clone dedicated demodulators. Which is still having to edit the source files. maybe a lock-unlock option would do. 
Keep it at 15k but offer us the option to unlock the field maybe?

AlexandreRouma:
If you "need" it you're using it for something it's not meant to do. 20KHz stations are already rare (basically just china that doesn't care about the usual broadcast limits) and 60KHz is down right ridiculous. If you're already misusing the AM demodulator for something it's not meant to do, then changing the code shouldn't be a big obstacle.

spithash:
"ridiculous"?

Issue #603, opened 2022-01-22
Support multiple resource directories


Rua:
Currently the resource directory seems to default to `/usr/share/sdrpp`, and SDR++ will only read resources like bandplans from there. But this is a system-wide root-owned directory. Would it be possible to support multiple resource directories, so that both the system directory and `~/.config/sdrpp` are used? Then it would allow users to create their own custom bandplans without messing with the system files.

AlexandreRouma:
It's planned

PR #1214, opened 2023-10-22
Load custom bandplans

glv2:
Load custom bandplans (located in "$HOME/.config/sdrpp/bandplans/" on Unix-like systems) in addition to the default bandplans.


AlexandreRouma:
as the PR template says, open an issue requesting the feature instead.

glv2:
There's already an open issue about this (#603).

Out of curiosity, what's the point of not accepting patches?
Even if a patch doesn't do something as you would have done it, it gives something to start with, and you can modify it.

AlexandreRouma:
I have very little free time and it's always quicker to just directly do it how I'd prefer it.
I already have to deal with my own previous code, if I add someone else's code to the mix it's just gonna be a mess.

PR #1226, opened 2023-11-11
Fix JACK audio repeating bug

illegalcoding:
# Fix JACK audio repeating bug

#### This pull request fixes a bug where if you use native JACK _(It doesn't happen with pipewire-jack IIRC)_ and pause the SDR it keeps playing whatever was last in the buffer and creates a very unpleasant noise.

I have made a pull request for this issue before back in April, but that fix was less than ideal, as it kept clearing the buffer over and over again, after it was already cleared. While trying to find a better solution I discovered that the `doStop()` function in the audio_sink is seemingly not always called on pause(?). This PR adds that call.

Another issue with the previous PR was that it completely prevented any audio playback through the sink while the SDR was stopped, which was a planned feature with a sink rewrite. This PR shouldn't prevent that, though the solution for this is _slightly janky_, it just immediately restarts the sink after stopping it. This could cause issues if one was to manually route SDRPlusPlus, where it would lose that connection, but this is easily fixable using the QJackCtl's patchbay, making it automatically connect wherever it needs to.

#### If there's any issues that I'm blind to, like there being a very specific reason doStop() isn't called, please let me know.

### DIFF
--- a/core/src/gui/main_window.cpp
+++ b/core/src/gui/main_window.cpp
@@ -682,6 +682,8 @@ void MainWindow::setPlayState(bool _playing) {
         playing = false;
         onPlayStateChange.emit(false);
         sigpath::sourceManager.stop();
+        sigpath::sinkManager.stopStream("Radio");
+        sigpath::sinkManager.startStream("Radio");
         sigpath::iqFrontEnd.flushInputBuffer();
     }
 }


AlexandreRouma:
Were you drunk when you wrote this PR? Seriously...

You realize that:
1) Obviously doStop of the **sink** isn't called when stopping the **SDR**, otherwise you would not be able to output audio with a stopped SDR
2) You hard-coded an audio stream name when SDR++ can have any number of demodulators and thus audio stream. And there might not even be one named "radio" at all...

I've already said it multiple times that it's a bug in RtAudio that sends the same buffer over and over instead of sending silence when data isn't available. On all of its other backends this works properly.

illegalcoding:
Come on, let's be professional now, no need for that kind of language. I'm just trying to contribute to an open-source project I like, and I am still learning.

For number 1, with a basic configuration, it is not intuitive that the audio sink would not stop when you stop the SDR, which is collecting all the data to be demodulated and sent to the audio sink. Which use cases would this functionality be necessary for? I was under the impression that functionality for outputting audio with a stopped SDR would be coming with a sink rewrite, one which seemingly hasn't had any commits since August.

As for number 2, I wrote "Radio" since by default, sinkManager.getStreamNames() gave me a vector with 1 string, "Radio". Functionality for iterating through these and deciding what to do with them could easily be addded.

I get that it's a bug in RtAudio, but as long as that bug is still there, a workaround would be nice for JACK users.

Could you tell me what is supposed to happen when the SDR is stopped? Keep the audio sink open but not send any data to it? What is the advantage of that over stopping the audio sink, and starting it with the SDR? I'm not trying to be condescending here, just quite confused.

Please, change your attitude, and don't be condescending to those who are trying to help.

Issue #557, opened 2021-12-25
Need SDRPlusPlus/SDR++ to run on our M1 and other Apple silicon Macs! Help!


NH7L:
Lots of us really want SDRPlusPlus SDR++ to run on our M1 and other Apple silicon Macs. Not at all happy being compelled to use CubicSDR, which is a non-intuitive pain in the ass. Most of us non-engineers, non-coders grew up on SDR#. SDR++ is the answer to our dreams. Hope you can help us, Alexandre! Many thanks for all you've done to create this great program.

AlexandreRouma:
SDR++ already runs natively on M1 just fine.
I dont own a mac tho so I can only provide intel builds built by the CI
system. (The .app bundle is available on https://www.sdrpp.org/nightly).
I've recently been lent a remote M1 mac mini but i only use it for release,
so the builds arent up to date)

NH7L:
Thank you, Alexandre, for answering me, and so quickly.

I realize SDR++ can be made to run on my M1 Mac, but I have no
background or knowledge in software engineering, coding, etc., so that is
beyond me.

I guess my alternative is to install Windows OS on my Mac and then install
the Windows version of your excellent app. I have installed CubicSDR, but I
find it awkward and awful to use. It is totally anti-intuitive. I may have
to learn it, though. Or I could install your program on my Windows laptop.  I
dearly want to run SDR++ on my Mac, however.

Like so many others, I learned what little I know about SDR through using
SDR# on the ancient Windows machine I keep. i am an amateur radio
enthusiast, U.S. license NH7L, and in general, most of us Mac radio
amateurs find ourselves compelled to keep the hated Windows machines around
just for amateur radio purposes, since there is relatively little available
for Macs.

Your app appears to be so good that it would become THE default SDR app for
Macs if it were easier to install. There are thousands and thousands of us!

Best wishes, merry Christmas and happy New Year.

Ernest Murphy NH7L
Birch Bay, Washington state, USA

https://www.qrz.com/db/nh7l

AlexandreRouma:
Unfortunately, making the build process on MacOS easier isn't possible, it's just the nature of software development.

[...]

PR #1247, opened 2023-12-06
Add MacOS security information

mahtin:
On modern MacOS's any downloaded app that is not signed (i.e. coming from the App Store etc) will be met with a security popup as MacOS refuses to run it. This can be overridden with a Ctrl-click, etc.

This PR updates the main README such that users can see this issue clearly. The edit adds minimum text and points to a correct comment in issue #557 by @michaelgeorge001.

Thank you for SDRPlusPlus!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -64,6 +64,8 @@ There are currently no existing packages for other distributions, for these syst
 
 Download the app bundle from the latest [nightly build](https://www.sdrpp.org/nightly)
 
+Note: On modern MacOS's the downloaded app will require (this)[https://github.com/AlexandreRouma/SDRPlusPlus/issues/557#issuecomment-1407393061] in order to successfully run. I.e. Control-click the app icon, then choose Open to override the security warning.
+
 ## BSD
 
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
I don't think this is something that ought to be in the documentation. It's the case for all unsigned software, not something specific to SDR++. It's the same as telling Windows Defender on Windows 10 to let run the file because it comes from the internet.

mahtin:
@AlexandreRouma - based on what you wrote in your comment; I'd go back and add that Windows10 text of yours to this page. I'd also add a Linux note saying "it just works". 

The point of documentation is to help the user at the point when help is needed. Being verbose isn't a bad thing! 

Hopefully you'll agree. Yes?

AlexandreRouma:
> The point of documentation is to help the user at the point when help is needed. Being verbose isn't a bad thing!

The issue is not verbosity, it's relevance. By that same logic the documentation would teach you how to extract a zip file because it's technically required to install the software. There is a limit to what is to be discussed in documentation, and if it's not strictly related to the software, it doesn't have a place in its documentation. Basic knowledge of the operating system is to be assumed.

mahtin:
> Basic knowledge of the operating system is to be assumed.

Clearly you've not been to our ham club meets! Hi hi!

PR #1267, opened 2024-01-04
add additional library to debian install instructions

clouserw:
I tried installing on a fresh install of Ubuntu 22 and `dpkg -i` failed due to `libzstd-dev` not being installed.  After installing it worked fine.


AlexandreRouma:
Can you move it to just before `libsoapysdr-dev `, I prefer all core dependencies to come  first before the module dependencies.

clouserw:
sure

PR #1284, opened 2024-01-21
Make compatible with rtaudio API 5 and 6.

hzeller:
Recent rtaudio changed the API to not throw exceptions anymore and also have DeviceIDs not queried by index but IDs that are provided separately ( https://github.com/thestk/rtaudio/releases ).

Adapt the code-base to be compatible with the old and the new API as we have to exepect that in a transition period both APIs are common on various build platforms.

### DIFF
--- a/sink_modules/audio_sink/src/main.cpp
+++ b/sink_modules/audio_sink/src/main.cpp
@@ -31,6 +31,10 @@ public:
         monoPacker.init(&s2m.out, 512);
         stereoPacker.init(_stream->sinkOut, 512);
 
+#if RTAUDIO_VERSION_MAJOR >= 6
+        audio.setErrorCallback(&reportErrorsAsException);
+#endif
+
         bool created = false;
         std::string device = "";
         config.acquire();
@@ -42,12 +46,15 @@ public:
         device = config.conf[_streamName]["device"];
         config.release(created);
 
-        int count = audio.getDeviceCount();
         RtAudio::DeviceInfo info;
+#if RTAUDIO_VERSION_MAJOR >= 6
+        for (int i : audio.getDeviceIds()) {
+#else
+        int count = audio.getDeviceCount();
         for (int i = 0; i < count; i++) {
+#endif
             try {
                 info = audio.getDeviceInfo(i);
-                if (!info.probed) { continue; }
                 if (info.outputChannels == 0) { continue; }
                 if (info.isDefaultOutput) { defaultDevId = devList.size(); }
                 devList.push_back(info);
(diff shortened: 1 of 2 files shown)


AlexandreRouma:
lines 34 -> 48: Omitting the triple line comment stating the obvious, having a lambda like that is just making the code more visually bloated. It must be moved to a private member function instead. Additially, looking at the error types, throwing just because something is a warning sounds like a bad idea:
image
This will result in a complete crash of the entire application if any of these gets fired outside of the trycatch, which considering it's a callback is not clear whether it's a possibility or not. Just print an error instead.

lines 58 -> 67: If you're gonna need an if macro anyway, why waste lines, memory and compute power generating a new array when you can just change the for statement itself... Also the entire code is `camelCase` so the variable name wasn't conform.

hzeller:
`Lines 34-> 48` : Done.
   * Added static member function and passed that to the `setErrorCallback()`
   * Made the choice of things to warn about and things to throw about a switch/case

`lines 58->67`: Done. Made the choice of source directly an `#ifdef` around the loop header.

AlexandreRouma:
Ok. Also don't remove the `info.probed` indiscriminately. I didn't just put them there to add more lines. On some system, not checking for that causes a crash because RtAudio does not guarantee all indices to be valid devices.

AlexandreRouma:
Other things while I'm at it:
- The `:` 4 characters away from each other are ugly, use ´flog::error("AudioSinkModule Error getting audio device ({}) info: {}", i, e.what());´ (yes the numbers in the log escape are useless, they were still there from the previous logger)
- `flog::warn("AudioSink: {0} ({1})", errorText, (int)type);` should be `flog::warn("AudioSinkModule Warning: {0} ({1})", errorText, (int)type);` for consistency

AlexandreRouma:
And also rename `reportErrorsAsException` to something more sensible and descriptive like `errorCallback`

hzeller:
Ok, I can re-add the `probed` for the old API.

I have a question in the AudioSource: 

https://github.com/AlexandreRouma/SDRPlusPlus/blob/fa76b4e865aeb6cebf0340bf135cd049b426c4f0/source_modules/audio_source/src/main.cpp#L93

The `info.probed` is used in a context where stereo channels are queried with the result that _non_-probed inputs that are non-stereo will pass through. This looks like it was meant to skip if this was not `info.probed`; so similar like in the AudioOutput `if (!info.probed) { continue; }` ?

AlexandreRouma:
Looks like a typo, was meant to be `||` instead of `&&`, you can switch it to that and it'll have the desired behavior

hzeller:
Addressed
  * `info.probed` is now used when the old API is active.
  * fix the warning and error log formatting as suggested.
  * changed `reportErrorAsException()` to `errorCallback()`

PR #1285, opened 2024-01-21
Make it easier to contribute: add .editorconfig

hzeller:
http://editorconfig.org/
The .editorconfig, which is a common way to tell IDEs and editors basic expectations such as indentation of code. Almost every editor supports that and people have it configured.

Set the configuration to the observed 4 space indent for c++ code.

That way, code contributions will automatically adhere to the basic formatting even if not evaluating .clang-format.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+# Editor config file, see http://editorconfig.org/
+root = true
+
+[*]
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
Any editor figures out all of this automatically so this is unnecessary bloat.

PR #1308, opened 2024-01-30
Added speedup ccache compilation

GermanAizek:
# Important
More information about package: https://ccache.dev


AlexandreRouma:
Code pull requests are not welcome, open an issue instead.

PR #1352, opened 2024-03-07
Fixed spectran_http_source

hb9fxq:
Fixed spectran_http_source block to work poperly with the IQ Demodulator Block, as intended by the Aaronia RTSA Suite HTTP API design.

Also introduced fixed / realistic sample rates to be used withing a network environment.

### DIFF
--- a/source_modules/spectran_http_source/src/main.cpp
+++ b/source_modules/spectran_http_source/src/main.cpp
@@ -68,6 +68,11 @@ public:
             strcpy(hostname, hostStr.c_str());
         }
 
+        if (config.conf[name].contains("demodulatorBlockApiName")) {
+            std::string demodulatorBlockApiNameStr = config.conf[name]["demodulatorBlockApiName"] ;
+            strcpy(demodulatorBlockApiName, demodulatorBlockApiNameStr.c_str());
+        }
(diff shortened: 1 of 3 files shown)


hb9fxq:
we'll put a documentation how to configure the mission with examples into the Aaronia Support forum. 

image

AlexandreRouma:
The code style doesn't match the rest of modern SDR++, some lines are commented out instead of removed (one of which shouldn't be commented in the first place), the ImGui widget IDs use the wrong prefix of `spectran` instead of `spectran_http` which will cause conflicts, the 4 commits are all named the same, the layout of the widgets is not done properly and their order is inconsistent with all other networked SDR++ sources.
image

Please, there's a reason the issue template and contributing.md says I don't accept pull request. Just opening an issue saying you wanted certain changes done to how it works would have been faster and easier for both of us. I appreciated that you want to directly contribute instead of having me do the work but in my situation it's just much me convenient for me to dit it right the first time around.

So please open an issue with what needs to change from the user's point of view and I'll implement something.

hb9fxq:
https://github.com/AlexandreRouma/SDRPlusPlus/issues/1354

PR #1418, opened 2024-06-09
Update france.json

Armand31:
Adding the 2 DAB bands and Police band (tetrapol protocol)


AlexandreRouma:
There are numerous JSON syntax errors, please at least check that what you submit is valid.

Armand31:
Sorry, fixed

PR #1476, opened 2024-09-14
Changes related with Idea on #1475

Zaryob:
It now adds bookmark with right click on frequency area. It makes easier to add bookmarks. Also PopUp changed with ImguiPopModal because if you not complate adding bookmark it continiously track you. And "Enter" "Escape" buttons added for organic input strategy. (It means I'm reflexively pressing enter to save. it is organic input strategy, if you learn you will use...)


AlexandreRouma:
As mentioned in the pull request template and contributing.md, code pull requests are not accepted.

PR #1501, opened 2024-10-22
Add Republic of Korea BandPlan

bsy0317:
Add Republic of Korea BandPlan.
It's not the Democratic People's Republic of Korea!


AlexandreRouma:
SDR++ doesn't support displaying characters outside the latin alphabet so thi bandplan won't display properly. Also there are some typos in the band types (eg. `marin` instead of `marine`)

bsy0317:
Hello! I've made the requested changes and submitted a pull request. Please review it. :)

AlexandreRouma:
Alright, just one last thing. The indentation has to be 4 spaces like it was before.

bsy0317:
Thank you for your quick response :)

bsy0317:
corrected that

bsy0317:
In addition, this is designated by the Korea Radio Administration

PR #1518, opened 2024-11-07
Fix ireland.json

Oskar-Dudek:
I used jsonlint.com and i edited it in some places and it says json valid, can you please try this version? Let me know if its broken or not

### DIFF
--- a/root/res/bandplans/ireland.json
+++ b/root/res/bandplans/ireland.json
@@ -24,6 +24,12 @@
             "end": 1602000
         },
         {
+            "name": "160m ham band",
+            "type": "amateur",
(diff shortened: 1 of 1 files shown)


AlexandreRouma:
Ok,
By the way, there is no need to open an entirely new PR every time you fix something in an existing PR, you can just edit the file and it will add to this one.

Oskar-Dudek:
so if i add something new i use this PR?

AlexandreRouma:
if I tell you to correct something while the pull request is still open (not merged), you can edit the file in your fork directly and it will show up here.

I'd recommend to get familiar with how git and github work.

PR #1613, opened 2025-04-04
Fix Fobos always using 50MHz

stenn930:
Looks like a clear typo/bug to me


AlexandreRouma:
Not only is not a bug, but it's very much on purpose to work around a hardware flaw of the FobosSDR where the antialiasing filter in front of the ADC is much too wide. Running at lower samplerates than 50MHz WILL cause aliasing.

PR #1678, opened 2025-08-21
Update bands based on the PNRF 2022

Fale:
# Important

Only bandplan, colormaps and themes are accepted. Code pull requests are **NOT welcome**. 
Open an issue requesting a feature or discussing a possible bugfix instead.


AlexandreRouma:
Do you have a source to verify this?
I see a lot of changes to frequencies which should only rarely happen...

PR #1690, opened 2025-09-27
Use brew volk package on MacOS instead of source build

dh1tw:
Recent versions of cmake require `cmake_minimum_required(VERSION)` to be higher than 3.5. Almost all CMakeLists.txt are already at VERSION=3.12. I found two more CMakeLists.txt where `cmake_minimum_required` had to be bumped to 3.12.

Volk can now be installed on MacOS with the Homebrew package manager. No more manual compiling needed. I updated the readme accordingly.

### DIFF
--- a/core/libcorrect/CMakeLists.txt
+++ b/core/libcorrect/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.12)
 project(Correct C)
 include(CheckLibraryExists)
 include(CheckIncludeFiles)
--- a/core/libcorrect/CMakeLists.txt
+++ b/core/libcorrect/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 2.8)
 project(Correct C)
 include(CheckLibraryExists)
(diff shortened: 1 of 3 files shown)


AlexandreRouma:
> Recent versions of cmake require cmake_minimum_required(VERSION) to be higher than 3.5. Almost all CMakeLists.txt are already at VERSION=3.12. I found two more CMakeLists.txt where cmake_minimum_required had to be bumped to 3.12.

Please PR to upstream instead:
* https://github.com/quiet/libcorrect
* https://github.com/discord/discord-rpc

> Volk can now be installed on MacOS with the Homebrew package manager. No more manual compiling needed. I updated the readme accordingly.

I can't test right now as I don't have a mac to validate that this is the case. Please edit the PR to only keep this change and I'll only merge once I can check first hand that the brew version works properly.

dh1tw:
Thanks for the swift feedback. I just opened the PRs as suggested in the upstream repos, but I doubt that they will ever get merged. There hasn't been any activity in these repos for more than 5 years.

Sure, I can remove the edit to the CMakeLists.txt. 

Would you prefer to add note in the readme.md that for modern versions of cmake, the flag `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` needs to be set? 

Here is the link to the [volk homebrew formula](https://formulae.brew.sh/formula/volk).

